PR #580, opened 2024-08-26
Correct README to clone from Github repo

MattDodsonEnglish:
Also, it seems that the references to the "private Gitlab repo" are no longer needed. Unless I've missed something, I was able to clone the repo and Kokkos sub-modules  without needing other access.


jmstone:
Thanks for pointing this out.  I've rewritten the README now so this PR has been superseded, and hopefully the updated version addresses this issue.

PR #583, opened 2024-09-02
hotfixing z4c amr and remove most warnings at compile time

HengruiPrinceton:
Z4c with AMR is broken in current master due to incorrect size of the T_\mu\nu array when AMR is invoked. This merge request fixes this. Also removes most of the warnings during compile.


jfields7:
I see a lot of changes other than in `tmunu.cpp`. I assume these are mostly for fixing warnings and removing dead code?

jfields7 (approved):

dradice (approved):

PR #586, opened 2024-09-07
add a new imex integrator

wisYue:


jmstone:
Thanks, Yue.  I've created a new branch in which I implemented the C-shock test, and added this integrator to that branch, so it will get added to main when I merge that branch.  So this PR is now not needed.

I'm afraid this new integrator does not pass the C-shock test, so we will have to look into this further before using it in applications.

wisYue:
Hi Jim,
  

  
What is the issue of the C-shock test? I will back to IAS next Monday, so probably we can talk more about this.
  

  
Best,
  
Yue
  
  
  
  
  
  
  

  
  
>

PR #588, opened 2024-09-09
Cleaned up and extended AMR logic for Z4c

dradice:
Redesign of the Z4c AMR to support BHNS and NSNS binaries
    
* PunctureTracker has been replaced by the more general CompactObjectTracker
* The user can now specify the minimum resolution at different radii
* The L2, LInf refinement criteria have been simplified and generalized
* CompactObjectTracker data and Z4c::last_output_time are added to the restart files
* All MPI_Reduces in WaveExtr have been consolidated into one


jfields7:
It looks like the position update in `CompactObjectTracker` is effectively done with an Euler step. Since the shift vector and fluid velocity are typically updated with something like RK3 or RK4, do we anticipate there being any issues with the tracker becoming desynched from the actual compact object?

dradice:
I think this was checked at some point with Cactus, but it was found that using higher-order integration does not improve the accuracy of the trackers in a meaningful way. For particles, I experimented with Adam-Bashforth in the past, but I found that it was not worth the extra cost and the reduced CFL. Note that Euler is also the only available method for the tracker in `main`, so this patch does not change that.

jfields7:
Okay. I assumed that was probably the case, but I thought I should ask first.

jfields7 (approved):

HengruiPrinceton (approved):
Able to reproduce waveform from headon bbh. Ready to be merged @jmstone 

image

jmstone:
> Able to reproduce waveform from headon bbh. Ready to be merged @jmstone

It seems this PR is still failing both the CPU and GPU regression tests with compile errors , so as soon as that is fixed I can merge.

HengruiPrinceton:
> > Able to reproduce waveform from headon bbh. Ready to be merged @jmstone
> 
> It seems this PR is still failing both the CPU and GPU regression tests with compile errors , so as soon as that is fixed I can merge.

Apology introduced a bug while fixing conflicts. Now everything is fixed!

PR #589, opened 2024-09-09
Bump Kokkos to 4.4

jmstone:
Time to bump Kokkos to the latest version again.  This enables getting rid of the patch required to fix the shared memory bug in Kokkos 4.1 as well.  The Newtonian solvers seem to work fine with this version on both CPUs and an A100, with no loss of performance.


HengruiPrinceton (approved):
Able to run z4c without performance change on A100

jfields7:
This update doesn't seem to be to the Kokkos 4.4.00 tag on the master branch but rather to a recent version of the Kokkos development branch. Is this actually what we want?

jmstone:
> This update doesn't seem to be to the Kokkos 4.4.00 tag on the master branch but rather to a recent version of the Kokkos development branch. Is this actually what we want?

No, we want the Kokkos 4.4.00 tag.  I would worry that things are less stable on the development branch.  My bad.  Could you commit the change that guarantees we get the 4.4.00 tag to this branch, and then I can merge it?  I am not expert at managing submodules.

jfields7:
This pull request also appears to be *into* `bump_to_kokkos4.4` *from* `main` rather than the other way around. I assume we actually want it the other way around?

jmstone:
> This pull request also appears to be _into_ `bump_to_kokkos4.4` _from_ `main` rather than the other way around. I assume this is not correct?

Yes, please fix my mess of a PR.  Thank you.

PR #592, opened 2024-09-13
Bump to kokkos4.4

jfields7:
This is a fixed PR request to replace #589, which was trying to merge `main` into `bump_to_kokkos4.4` rather than the other way around.


felker:
This should probably be bumped to 4.4.01? Just released on the 13th: https://github.com/kokkos/kokkos/releases/tag/4.4.01

PR #602, opened 2024-09-23
Make restart file compatible with previous version

mh-guo:
The `RegionSize` struct has been changed since the merge of the DynGRMHD module. This leads to a change in the structure of the restart file so the the code cannot read the previous restart file correctly. This PR rollbacks the `RegionSize` struct so that the restart file is compatible with previous versions. In addition, the new variables `ixd1, idx2, idx3` in `RegionSize` are used only once in the code so it seems not necessary. I'm also open to keeping these variables if they are necessary.

### DIFF
--- a/src/dyn_grmhd/dyn_grmhd.cpp
+++ b/src/dyn_grmhd/dyn_grmhd.cpp
@@ -474,7 +474,7 @@ void DynGRMHDPS<EOSPolicy, ErrorPolicy>::AddCoordTermsEOS(const DvceArray5D<Real
                     &g3u[S11], &g3u[S12], &g3u[S13], &g3u[S22], &g3u[S23], &g3u[S33]);
 
     // Calculate the metric derivatives
-    Real idx[] = {size.d_view(m).idx1, size.d_view(m).idx2, size.d_view(m).idx3};
+    Real idx[] = {1./size.d_view(m).dx1, 1./size.d_view(m).dx2, 1./size.d_view(m).dx3};
     Real dalpha_d[3] = {0.};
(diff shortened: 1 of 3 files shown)


jfields7:
I don't remember exactly why I did this. It might have been to match GR-Athena++, or perhaps it was to eliminate the extra divisions. Nevertheless, it's a little surprising that this breaks the restart files; we've had issues with other quantities in the code needing to be added explicitly to the checkpointing procedure. Is the mesh saved via a straight memory dump into the file?

mh-guo:
Yes the mesh is directly dumped into the restart file in `restart.cpp`, line 202,
`    resfile.Write_any_type(&(pm->mesh_size), (sizeof(RegionSize)), "byte");
`
So a change of `RegionSize` leads to a change of the restart file that the code cannot handle.
For other quantities in the code needing to be added explicitly to the checkpointing procedure, the code would check, say, whether `pz4c` is constructed, so it can handle them properly.
Which solution do you think is better?

mh-guo:
In addition, it's really the size of MeshBlock,  `mb_size`, not the Mesh size `mesh_size` that uses the new variables like `idx1`. So the new variables in `mesh_size` are never used. Perhaps we could treat them separately instead of using the same struct `RegionSize`?

jfields7:
I'm fine with getting rid of it. I ran some quick tests comparing performance. It might be marginally faster to store the reciprocal spacings rather than doing the divisions on the fly, but it's not worth the headache of breaking everyone else's restart files.

jfields7 (approved):

jmstone (approved):

PR #607, opened 2024-10-03
Output changes for DynGRMHD

PeterHammond:
Add output and storage for temperature. Rename `eint` to `press` when dynamical GRMHD is active.


src/outputs/basetype_output.cpp
```diff
@@ -451,6 +467,13 @@ BaseTypeOutput::BaseTypeOutput(ParameterInput *pin, Mesh *pm, OutputParameters o
       outvars.emplace_back("bcc3",2,&(pm->pmb_pack->pmhd->bcc0));
     }
 
+    // MHD temperature
+    if (variable.compare("mhd_t") == 0 ||
+        ((variable.compare("hydro_w") == 0 ||
```
jfields7, line 472: This should be `mhd_w`, not `hydro_w`.

jfields7 (changes requested):

src/outputs/basetype_output.cpp
PeterHammond: Done.

PR #610, opened 2024-10-08
Time dependence and AMR for ADM

### LINKED ISSUE #600: AMR does not work with DynGRMHD in fixed spacetimes
The `DynGRMHD` solver can be used in fixed spacetimes if `<adm>` but not `<z4c>` is enabled in the parameter file. This is very useful for problems which need a generic but fixed spacetime, such as the torus test, various TOV problems, etc. However, the ADM variables used to define the spacetime are not evolved variables; they are defined in the problem generator, and currently the only way to update them is via `Z4cToADM`, which will never be called if Z4c is not also enabled. Consequently, only SMR is supported for fixed generic spacetimes.

This isn't a major issue at the moment, but there are a handful of interesting problems where a fixed (or analytic) spacetime could benefit from AMR. One possible solution is to have an optional `UpdateADM` function specified by the problem generator which could be used to update the ADM variables during remeshing operations. This functionality could also be generalized to allow for specifying time-dependent but analytically specified metrics regardless of whether or not AMR is enabled, which is something that a few people are interested in doing with AthenaK.

jfields7:
This merge request adds a new function pointer, `SetADMVariables`, to the `ADM` class. By default it uses `ADM::SetADMVariablesToKerrSchild`, which uses the existing Kerr-Schild infrastructure. By setting `coord/minkowski = true` in the parameter file, this reverts to flat Minkowski space as it would for anything else using the Kerr-Schild functions. This has the added benefit that many of the shared pgens using Minkowski or Kerr-Schild coordinates are cleaned up and replaced with a single call to `padm->SetADMVariables()` instead.

This API is intended to fix #600. When AMR is enabled, `SetADMVariables` is called at the end of `MeshRefinement::RedistAndRefineMeshBlocks`, which means that the ADM metric will be completely defined everywhere on the new grid, allowing AMR evolution with `DynGRMHD` without needing to enable Z4c as well. Other metrics can be defined by pointing `SetADMVariables` to a different function; it should be a void function that takes the `MeshBlockPack` as a single argument.

An additional task has been added to the `DynGRMHD` task list to allow for *dynamical* ADM evolution, and examples are contained in an extension to the `blast` pgen with an FLRW metric and the new `dynbbh` pgen, which superposes two Kerr-Schild black holes and moves them along a circular orbit. Dynamical evolution in the task list can be enabled with `adm/is_dynamic = true`.

# Draft to-do list:
- [ ] ~~Update TOV pgen to use `SetADMVariables`.~~ This probably needs to be postponed and done as part of the TOV refactor.
- [x] Check that all pgens compile and work as intended.
- [x] Check that GRMHD+Z4c still works as intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@ Using Kokkos enables *performance-portability*.  AthenaK will run on any hardwar
 AthenaK is targeting challenging problems that require exascale resources, and as such it does not implement all of the features of Athena++.  Current code features are:
 - Block-based AMR with dynamical execution via a task list
 - Non-relativistic (Newtonian) hydrodynamics and MHD
-- Special relativistic hydrodynamics and MHD
-- General relativistic hydrodynamics and MHD in stationary spacetimes
+- Special relativistic (SR) hydrodynamics and MHD
+- General relativistic (GR) hydrodynamics and MHD in stationary spacetimes
 - Relativistic radiation transport
 - Lagrangian tracer particles, and charged test particles
 - Numerical relativity solver using the Z4c formalism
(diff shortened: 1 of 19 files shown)


HengruiZhu99:
@jfields7 @jmstone I'm getting the following compilation error on della because of the change in shock_tube.cpp; have you guys seen this?

/home/hz0693/athenak/src/pgen/tests/shock_tube.cpp(129): error: identifier "_NV_ANON_NAMESPACE::shk_dir" is undefined in device code

jfields7:
It seems to me that I saw this once before, but apparently it didn't end up in this branch. I think it's a very selective issue that only shows up in some versions of the CUDA compiler. If I recall correctly, it's related to the variable `shk_dir` declared inside the anonymous namespace at the beginning of the pgen file. The easiest solution is to replace the references on lines 125 and 331 `auto &shk_dir_ = shk_dir` with the direct copies `auto shk_dir_ = shk_dir`.

PR #611, opened 2024-10-15
Update CMakeLists.txt with bin install

VictorEijkhout:
The executable was not being installed to the bin directory.
I added a `install` line.


jmstone:
Victor: thanks for the contribution.  I am a cmake newbie.  Where is the bin directory located?  Does cmake create it automatically? When I run cmake with this change on my laptop the executable still appears in build/src/

PR #613, opened 2024-10-30
uplifting z4c performance

HengruiZhu99:
Work done with @jfields7 during the TACC workshop to optimize the performance of the z4c module. By eliminating the index map and making proper use of the tensor symmetry, the register spillage is more than halved, and we achieved a 50% improvement in the overall zone-cycles-per-second.

We also find that the order of operation in the cuda kernel does matter for performance. Due to the complexity of inter-dependencies of the intermediate variables, we will optimize that further, which may require reusing variables that are defined on the stack. 

I also have a bbh calculation running to make sure the waveform is still consistent with our previous calculations, hence currently draft. 

Performance comparison with current main, using the z4c_one_puncture pgen with SMR:
A100: 1.292741e+07->1.887377e+07
GH200: 2.357512e+07->3.528358e+07

Able to reproduce waveform from head-on collision:
image


jfields7:
A few comments that can be addressed while this is still a draft:
1. You can remove all my commented-out code in `athena_tensor.hpp`.
2. I don't think `z4c\z4c_punctures.cpp` is actually used anywhere in the code; it doesn't show up in the `CMakeLists.txt` file, and it looks so outdated I don't think it would compile if it were. Should this just be removed?
3. What do the changes to the linting script do?

HengruiZhu99:
> A few comments:
> 
> 1. You can remove all my commented-out code in `athena_tensor.hpp`.

Done!

> 2. I don't think `z4c\z4c_punctures.cpp` is actually used anywhere in the code; it doesn't show up in the `CMakeLists.txt` file, and it looks so outdated I don't think it would compile if it were. Should this just be removed?

Done!

> 3. What do the changes to the linting script do?

I'm removing athena_tensor.hpp from linting as somehow it complains about the templated const_expr for setting the static array size. I tried several things but cannot get around that. Do you have any suggestions?

jfields7:
> I'm removing athena_tensor.hpp from linting as somehow it complains about the templated const_expr for setting the static array size. I tried several things but cannot get around that. Do you have any suggestions?

You can add `// NOLINT` to the end of the offending lines. The `cpplint.py` script is not particularly bright and has a number of problems other than this.

HengruiZhu99:
> > I'm removing athena_tensor.hpp from linting as somehow it complains about the templated const_expr for setting the static array size. I tried several things but cannot get around that. Do you have any suggestions?
> 
> You can add `// NOLINT` to the end of the offending lines. The `cpplint.py` script is not particularly bright and has a number of problems other than this.

Done! Thanks for the suggestion!

jfields7 (approved):

PR #616, opened 2024-11-16
horizon find and cartesian grid dump

HengruiZhu99:
dumping data on Cartesian grid with application to horizon finder

### DIFF
--- a/src/z4c/compact_object_tracker.hpp
+++ b/src/z4c/compact_object_tracker.hpp
@@ -55,11 +55,11 @@ class CompactObjectTracker {
   inline Real GetRadius() const {
     return radius;
   }
+  Real pos[NDIM];
 
  private:
   bool owns_compact_object;
   CompactObjectType type;
-  Real pos[NDIM];
   Real vel[NDIM];
   int reflevel;         // requested minimum refinement level (-1 for infinity)
   Real radius;          // nominal radius of the object (for the AMR driver)
--- a/src/z4c/compact_object_tracker.hpp
+++ b/src/z4c/compact_object_tracker.hpp
@@ -55,7 +55,6 @@ class CompactObjectTracker {
   inline Real GetRadius() const {
     return radius;
   }
-  Real pos[NDIM];
 
  private:
   bool owns_compact_object;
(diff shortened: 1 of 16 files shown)


src/z4c/compact_object_tracker.hpp
dradice, line 1: It would be better not to make pos public and instead use GetPos() to get the compact object position

src/z4c/horizon_dump.cpp
dradice, line 1: * I do not understand the logic for why `parfilename` is given as a parameter to the function and then overwritten. Also, there is no guarantee that `parfilename` will really be 100 characters wide, so it would be safer to pass a `char const *` if you want to have the filename as input.
* You should not use `sprintf`, use `snprintf` instead (or use C++ strings)

src/z4c/z4c.cpp
dradice, line 1: Why are you using `make_unique` instead of `emplace_back`? I think it is better to have a list of objects, rather than a list of pointers

dradice (commented):
I am not sure why we would want to convert the tracker list to a list of pointers, which is what is being done here.

src/z4c/compact_object_tracker.hpp
HengruiZhu99: > It would be better not to make pos public and instead use GetPos() to get the compact object position

Done!

src/z4c/horizon_dump.cpp
HengruiZhu99: Fixed! Originally I wanted to run the ahfinder as soon as the file is dumped, so I'm saving the name of the parfile. But I then decided against that.

src/z4c/z4c.cpp
HengruiZhu99: I'm using the smart pointer to avoid complications of memory management using std::vector

HengruiZhu99:
> I am not sure why we would want to convert the tracker list to a list of pointers, which is what is being done here.

I mainly want to use the std::vector because of the easiness of accessing the elements. For the std::list I need an iterator making it more lengthy to pass the position of puncture into the horizon dump object.

PR #620, opened 2024-12-05
Fix "_NV_ANON_NAMESPACE::shk_dir" error.

jfields7:
This should fix the issue @HengruiZhu99 encountered while compiling on Della with the most recent changes. I encountered a similar issue on Vista. It appears to be due to a variable in the anonymous namespace not being captured properly by the GPU. Copying the offending variable rather than trying to use a reference appears to fix the issue. It isn't clear why it only appears with some versions of the Nvidia compiler.


jfields7:
> Hey Jacob, I've included these in the CCE merge request as well! Thanks for the suggestions!

Great, I'll go ahead and close this request, then.

PR #625, opened 2025-01-16
add two apple-with-apple tests

HengruiZhu99:
Adding robust stability and the gauge wave test. As well as fixing a few bugs, including mismatched variable name and index for the z4c user history output. 

Results are attached below:

Minkolski + noise, damping away as expected
stability

1d gauge wave with different order of spatial differencing + RK4, seeing expected order of convergence except at high resolution with high-order methods where the round-off error due to the increased numbers of operations now dominates, giving a negative convergence factor. 

gauge_wave


jmstone (approved):
All looks good, just have to decide in they should be in /pgen or pgen/tests

HengruiZhu99:
@jmstone moved z4c apple-with-apple tests out of the test folders, now ready to be merged!

PR #627, opened 2025-01-25
Cargrid output

dradice:
`CartesianGridOutput` can output any variable in a Cartesian box (within the simulation domain) at a prescribed position and extent.
    
To use it, add a block to the output file like this:
    
```
            <output3>
            file_type  = cart
            variable   = hydro_w
            dt         = 0.1
            center_x   = 0.0
            center_y   = 0.0
            center_z   = 0.0
            extent_x   = 0.1
            extent_y   = 0.1
            extent_z   = 0.2
            chebyshev  = 1      # 0:uniform grid; 1:Chebyshev-Lobatto grid
 ```
    
Note that the extent of the grid is actually twice of the extent variable. This is because we are using the same convention as in `CartesianGrid`.
    
This commit also adds `vis/python/cartgrid.py` to read the data produced by `CartesianGridOutput`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -88,6 +88,7 @@ add_executable(
         outputs/io_wrapper.cpp
         outputs/outputs.cpp
         outputs/basetype_output.cpp
+        outputs/cartgrid.cpp
         outputs/derived_variables.cpp
         outputs/binary.cpp
         outputs/eventlog.cpp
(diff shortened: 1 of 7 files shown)


HengruiZhu99:
Thanks for catching the bug in AMR. I commented those out for debug earlier and forgot to remove those when committing to the repo

HengruiZhu99 (approved):

PR #629, opened 2025-02-01
Speed up the legendre root finding (used in gauss-legendre grid)

HengruiZhu99:
updating the bisection solver by a newton raphson finder as @dradice suggested. Now speed up the code by quite a bit. I also included a unit test for doing cross integration of spherical harmonics on Gauss-Legendre grid. The integral is calculated robustly up to l_max=20, loss of precision seen at l_max>=25, generating an error on the order of 1e-8 for higher l's.  This is quite enough for CCE but should be kept in mind for future usage. Perhaps using precomputed values for the legendre roots and weights, or using the gsl implementation would be more robust.

### DIFF
--- a/src/utils/legendre_roots.hpp
+++ b/src/utils/legendre_roots.hpp
@@ -1,127 +1,159 @@
 #ifndef UTILS_LEGENDRE_ROOTS_HPP_
 #define UTILS_LEGENDRE_ROOTS_HPP_
+
 //========================================================================================
 // AthenaXXX astrophysical plasma code
-// Copyright(C) 2020 James M. Stone <jmstone@ias.edu> and the Athena code team
+// Copyright(C) 2020 James M. Stone <jmstone@ias.edu>
 // Licensed under the 3-clause BSD License (the "LICENSE")
 //========================================================================================
 //! \file legendre_roots.hpp
-//! \brief find roots of legendre polynomial
+//! \brief find roots of legendre polynomial using Newton–Raphson with Chebyshev guesses
 
 #include <iostream>
 #include <cmath>
-#include <list>
 #include <vector>
 
 #include "athena.hpp"
 
-/*Legendre Polynomial P0(x)*/
-double P0(double x) {
-  return 1;
+//-------------------------------------------------------------------------
+// 1. Basic Legendre polynomials
+//-------------------------------------------------------------------------
+/* P0(x) */
+inline double P0(double x) {
+  return 1.0;
 }
 
-/*Legendre Polynomial P1(x)*/
-double P1(double x) {
+/* P1(x) */
+inline double P1(double x) {
   return x;
 }
 
-/*Nth Legendre Polynomial Pn(x)*/
+/* Nth Legendre Polynomial, Pn(x) */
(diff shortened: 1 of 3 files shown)


src/utils/legendre_roots.hpp
```diff
@@ -1,127 +1,159 @@
...
-    if(root!=999) {
-      xi.push_back(root);
-      roots_weights[0][i]=root;
-      i++;
-    }
+  std::vector<double> xi(n);
+
+  // Parameters for Newton iteration
+  double tol      = 1e-14;
+  int    max_iter = 100000;
```
dradice, line 132: I think 100,000 iterations is way too many. You should be gaining 1 bit of precision at least per iteration and there are only 52 bits in a double precision number.

src/utils/legendre_roots.hpp
```diff
@@ -1,127 +1,159 @@
...
-      //(Modify the number of sub-intervals according to your needs)
-      roots_weights[1][i]=Ci(i,n,xi,-1,1,1000000);
+  // Compute weights (using the original Lagrange-based Ci or direct formula).
+  // Here we keep your original Lagrange-based approach:
+  for (int i = 0; i < n; i++) {
+    // Adjust N for integration as you see fit.
+    // For a large n, you might not need extremely large N
+    // if the function is well-behaved.
+    // But here we keep your original 60000 intervals.
+    roots_weights[1][i] = Ci(i, n, xi, -1.0, 1.0, 60000);
```
dradice, line 153: Use analytic expression for the roots weights, e.g., Canuto, Hussaini, Quarteroni, Zang, Eq. (2.3.10)

src/utils/legendre_roots.hpp
```diff
@@ -1,127 +1,159 @@
...
-
-    if(root!=999) {
-      xi.push_back(root);
-      roots_weights[0][i]=root;
-      i++;
-    }
+  std::vector<double> xi(n);
+
+  // Parameters for Newton iteration
+  double tol      = 1e-14;
```
dradice, line 131: I would iterate until floating point precision here. In my experience when $N \gtrsim 20$ the precision of the points become crucial for accuracy.

dradice (changes requested):
I have suggested a few more tweaks to improve accuracy and speed.

src/utils/legendre_roots.hpp
HengruiZhu99: Done! I left it the same as the N for bisection and forgot to update. The typical number of iteration to reach machine precision is 3, but I have set it to 20 for safety.

src/utils/legendre_roots.hpp
HengruiZhu99: Done. Thanks for the reference!

src/utils/legendre_roots.hpp
HengruiZhu99: Now set to std::numeric_limits<double>::epsilon()

dradice (approved):

HengruiZhu99:
@jmstone Ready to be merged!

PR #637, opened 2025-02-27
Remove team barriers in bvals_cc (Aurora fix)

michaelbuehlmann:
fixes deadlock on Aurora

Note from Daniel Arndt on why this wasn't an issue previously:
> For Cuda, threads that return from a kernel don’t participate in `__threadfence_block` calls which is not the case for SYCL. Also, it might quite well be that for Cuda we happen to execute the same number of work items per thread. Finally, calling team_barrier at the end of a kernel shouldn’t be necessary.

### DIFF
--- a/src/bvals/bvals_cc.cpp
+++ b/src/bvals/bvals_cc.cpp
@@ -115,14 +115,12 @@ TaskStatus MeshBoundaryValuesCC::PackAndSendCC(DvceArray5D<Real> &a,
             [&](const int i) {
               rbuf[dn].vars(dm, (i-il + ni*(j-jl + nj*(k-kl + nk*v))) ) = a(m,v,k,j,i);
             });
-            tmember.team_barrier();
           // if neighbor is at coarser level, load data from coarse_u0
           } else {
             Kokkos::parallel_for(Kokkos::ThreadVectorRange(tmember,il,iu+1),
             [&](const int i) {
               rbuf[dn].vars(dm, (i-il + ni*(j-jl + nj*(k-kl + nk*v))) ) = ca(m,v,k,j,i);
             });
-            tmember.team_barrier();
           }
 
         // else copy into send buffer for MPI communication below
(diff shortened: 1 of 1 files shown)


felker:
nice find! I have been puzzling over this the past two days, still trying to explain why the SYCL backend worked on Sunspot and pre-production Aurora last year (and on Sunspot even earlier than that, about 2 years ago, see https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenak/-/merge_requests/103 "Make bvals/ references to global host variable "my_rank" normal local variables")[^1]

I was using Kokkos 3.7.01 at my earliest successful runs on Sunspot, and I _believe_ 4.0.00 last year.

Looking at the git blame, many team barriers were added before that, e.g. in Nov 2022 (98eadc8862ec167265e219afeb84ed3423877fa5), but others _barely after_ my initial Sunspot tests (5315565)--- those latter barriers are the ones you removed in this PR. It is possible that I kept using an old AthenaK from before 5315565 even on Aurora until recently, which would explain the discrepancy

There are 130 `team_barrier()` calls in the code currently. How many of them are essential, and how many of them will cause issues with SYCL if the corresponding algorithmic or physics feature is enabled??

Will double check your removals with my hang reproducer on Aurora--- I was getting hangs even with a single MeshBlock (and of course no MPI) with the linear hydro wave. 


[^1]: an aside, it seems like many merge requests and issues did not get migrated from the original GitLab repo. A bit unfortunate since there is a lot of info contained within those--- maybe possible to transfer them, still?

felker:
and for my own records, the text of #103:
> The problems arising from this pattern are unique to host global variables. In C++ lambda functions, global variables (and references to global variables) behave as if captured by reference, but they are actually accessed directly as any non-lambda function would use global variables.
`KOKKOS_LAMBDA` gets redefined to some lambda with capture-default `[=]` capture by value, depending on the backend, so that the variables are defined on the device. But only variables with automatic storage duration declared within the "reaching scope" of the lambda are implicitly captured. So, a global variable defined on the host is not defined on the device nor captured by value. Plus, global (non-const) device variables are not directly supported in SYCL, you need `sycl::device_global` for that. 
An aside: there is some confusion about whether or not references to global variables should be implicitly captured by the language of the standard. GCC does, Clang does not: https://godbolt.org/z/aGbrEKb5e
That might explain why the code worked with the CUDA backend in the past, if you always used GCC and never Clang+CUDA.
All of these details are irrelevant to this application, since I doubt you would ever want a reference `int &my_rank` since it should never change anyway. But it was a fun C++ standard rabbit hole to explore.

jmstone:
This was superseded by #668

PR #639, opened 2025-03-04
Update AMR arguments in some hydro input files

yoonso0-0:
Hi everyone,

Here I am submitting a quick update on some input files, changing `refine_interval` to `refinement_interval` in the `<mesh_refinment>` block, which seems to be the correct argument name to be parsed. Please take a look!


jmstone:
Thanks for finding and fixing these!

PR #640, opened 2025-03-06
Add function to update particle information

DVilla95:
Fixes output bug where vtk had constant length even though particles leave the box over time.


jmstone:
Looks good, although I would suggest changing the function name to, e.g. CountParticles() since the current name is not very descriptive of what the function is actually doing.

PR #643, opened 2025-04-02
Added WatchDog to terminate hung simulations

dradice:
This commits ports the WatchDog implementation I wrote for the Einstein Toolkit [1] to AthenaK. This can be enabled using the CLI option "-w ss" with the timeout in seconds for the watchdog.

---
[1] https://bitbucket.org/cactuscode/cactusutils/src/master/WatchDog/


jmstone (commented):
Looks good.  I have not used pthreads explicitly, only OpenMP for managing multiple threads.  Is this adding an external library dependence beyond what is included in C++17?  It seems no, but I wanted to be sure.

dradice:
Technically I use POSIX threads, so this patch introduces a dependency on POSIX (Unix/Linux). In practice, the only widespread non-POSIX OS is Windows, but I think it would be difficult to port AthenaK to Windows in any case.

jmstone:
OK, it's been over a year -- finally time to merge this (thank Jacob for bugging me about it)....

PR #649, opened 2025-05-16
Fix compiler warnings in mesh.cpp

dradice:
This replaces four automatic arrays with dynamically allocated arrays.


jmstone:
This was closed with PR #756

PR #650, opened 2025-06-04
Fix CUDA compiler warnings observed on ALCF Polaris; bump Kokkos to 4.6.01

felker:
I believe these were all introduced to `main` within the last year--- it used to compile without warnings (default pgen option) on the machine. #613 added some warnings with the lack of `constexpr` fallthrough on the `enum`. Otherwise mostly standard "unused variable" and implicit capture of `this` in device lambdas. 

Still lots of CUDA and SYCL backend warnings and errors with some Z4c things and various problem generators that I didn't bother to fix (and some already addressed in other brances). E.g. `Kokkos::printf` should replace `printf` in https://github.com/IAS-Astrophysics/athenak/blob/cc65bf80d8d52f2034b6b194ab959a6ef4a732b5/src/pgen/sgrid_bns.cpp#L772-L783

Kokkos 4.6.01 removes a lot of deprecation warnings on ALCF Aurora and has been shown to be stable with Z4c tests.

### DIFF
--- a/src/athena_tensor.hpp
+++ b/src/athena_tensor.hpp
@@ -437,6 +437,7 @@ class AthenaPointTensor<T, sym, ndim, 3> {
         return data_[c + ndim*(a*(2*ndim - a + 1)/2 + b - a)];
       }
     }
+    return data_[c + ndim*(b + ndim*a)]; // Default to NONE case
   }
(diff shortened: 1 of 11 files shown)


jmstone:
This was superseded by #756

PR #659, opened 2025-07-23
Update z4c history and horizon finder defaults

HengruiZhu99:
A couple of updates:

1. z4c history variables:
    A. use the proper volume when calculating the L2 norm
    B. excise the puncture region based on the value of the conformal factor chi

2. horizon finder:
    A. change the spin from Killing (which is not working) to Axial
    B. change the surface interpolator to Hermite order 3 to improve stability
3. pgen for twopuncture:
    A. spectral interpolation instead of taylor expansion to interpolate data from twopuncture to athenak


src/outputs/history.cpp
```diff
@@ -197,18 +201,36 @@ void HistoryOutput::LoadZ4cHistoryData(HistoryData *pdata, Mesh *pm) {
...
     j += js;
 
-    Real vol = size.d_view(m).dx1*size.d_view(m).dx2*size.d_view(m).dx3;
+    Real detg = adm::SpatialDet(adm.g_dd(m,0,0,k,j,i), adm.g_dd(m,0,1,k,j,i),
+                                adm.g_dd(m,0,2,k,j,i), adm.g_dd(m,1,1,k,j,i),
+                                adm.g_dd(m,1,2,k,j,i), adm.g_dd(m,2,2,k,j,i));
 
-    // Hydro conserved variables:
+    Real vol = size.d_view(m).dx1*size.d_view(m).dx2*size.d_view(m).dx3
+               * std::pow(std::abs(detg),0.5);
```
dradice, line 209: Why are you using pow instead of sqrt?

src/outputs/history.cpp
```diff
@@ -197,18 +201,36 @@ void HistoryOutput::LoadZ4cHistoryData(HistoryData *pdata, Mesh *pm) {
...
     array_sum::GlobalSum hvars;
-    hvars.the_array[0] = vol*u_con_(m,0,k,j,i); // ||C||^2 (comes already squared)
-    hvars.the_array[1] = vol*SQR(u_con_(m,1,k,j,i)); //||H||^2
-    hvars.the_array[2] = vol*u_con_(m,2,k,j,i); // ||M||^2 (comes already squared)
-    hvars.the_array[3] = vol*u_con_(m,3,k,j,i); // ||Z||^2 (comes already squared)
-    hvars.the_array[4] = vol*SQR(u_con_(m,4,k,j,i));      // ||Mx||^2
-    hvars.the_array[5] = vol*SQR(u_con_(m,5,k,j,i));      // ||My||^2
-    hvars.the_array[6] = vol*SQR(u_con_(m,6,k,j,i));      // ||Mz||^2
-    hvars.the_array[7] = vol*SQR(u0_(m,I_Z4c_Theta_,k,j,i)); // ||Theta||^2
+    if (z4c.chi(m,k,j,i)>=0.0625) {
```
dradice, line 213: Can we have this be a parameter from the pgen, rather than a magic number in the code?

dradice (changes requested):
In the Z4c history, it would be better to have the user define the threshold for chi, rather than having it hardcoded.

src/outputs/history.cpp
HengruiZhu99: changed to sqrt now

PR #664, opened 2025-08-22
WENOZ reconstruction tweaks

ajdittmann:
I updated the comments in the WENO-Z reconstruction file to correctly represent which algorithms are currently employed by the code. 
I also cut out some of the redundant operations.


ajdittmann:
In principle, one could also rewrite each alpha calculation explicitly into a fused multiply-add, but I have no clue whether compilers will do that automatically with standard optimization settings or if the sliver of performance would be worth the loss of readability.

jmstone:
Looks good, thanks for the updates.

PR #665, opened 2025-08-26
Added evaluation of exit code to run_test_suite

davidvelasco07:


davidvelasco07 (commented):
Added evaluation of exit code to run_test_suite

davidvelasco07 (commented):
Added evaluation of exit code to run_test_suite

davidvelasco07 (commented):
It does capture fails in any of the 4 regression tests. I checked only with cpu, but it uses the same function on all cases.
You can ignore the last commit where I introduced an error in test_dyngrmhd_lwava1d_cpu to check that the error capture works.

PR #666, opened 2025-09-11
DynGRMHD updates: refactored TOV solver, tabulated EOS support, etc.

jfields7:
This pull request is a major update to `DynGRMHD`. Here's a summary of the biggest changes:
* The TOV solver has been separated from `dyngr_tov.cpp` and placed inside `utils/tov`. This is necessary because although `dyngr_tov.cpp` is the only included pgen that uses the TOV solver, many users have been writing separate TOV pgens for special problems (custom magnetic fields, custom source terms, spherical coordinates, etc.).
* New `TOVEOS` classes have been added to `utils/tov`, which are 1D EOSs that can be used for computing thermodynamic quantities for initial data (TOV, RNS, BNS, etc.).
* `EOSCompOSE` has been thoroughly tested and debugged. A unit test, `pgen/unit_tests/eos_compose_test.cpp`, has been added for testing purposes.
* To help with forthcoming support for neutrinos, `Primitive::EOS` now includes compile-time tags for whether or not entropy and chemical potential functions are available. This also means that the unused `Entropy` functions in `Primitive::IdealGas` and `Primitive::PiecewisePolytrope` have been removed, as they are no longer needed to compile correctly.
* Also looking forward to neutrinos, `Primitive::EOS` now has functions to compute beta equilibrium and trapped neutrinos.
* Support for 1D tabulated EOSs with gamma-law thermal extensions has been added with `Primitive::EOSHybrid`.
* Novel second-order "not-quite-transcendental" approximations to $\log_2 x$ and $2^x$ (see _ApJS_ 277 (2025) 2, 65, [2501.05410 [physics.comp-ph]](https://arxiv.org/abs/2501.05410)) have been added for interpolating tables, which can be enabled using `mhd/use_NQT=true`. These offer a modest performance boost for tabulated EOSs on compatible architectures.
* ~~The C2P operation for `DynGRMHD` now optionally supports initializing the root find with the solution from the last step. This can be enabled with `mhd/c2p_caching=true`. Experimentation suggests it does not usually improve performance, but it's included as an option.~~
* `DynGRMHD` now stores the temperature following the C2P for output purposes.
* `CompactObjectTracker` now supports a new tracking mode which does a nearest-neighbor search for the lowest value of $\alpha$ and snaps to that cell. Though it's not as smooth as advecting the trackers as is done in the BBH case, this tends to be more robust for BNS runs. It can be enabled with `z4c/tracker_mode=walk` for the new behavior and `z4c/tracker_mode=ode` for the old behavior.

Because this pull request has been delayed for so long and is already so massive, there are a number of other planned updates which have *not* been incorporated, including (but not limited to):
* Updating `lorene_bns.cpp` to read data using total energy density instead of baryon number density. This causes some small errors in the initial data for tabulated BNS runs, but fixing it requires some updates to the 1d TOVEOS infrastructure to work consistently for all EOSs.
* `dyngr_tov.cpp` does not currently support AMR when Z4c is disabled because it does not have a custom `SetADMVariables` function. This should not be difficult to do, but I don't necessarily think it's a priority because most TOV problems in fixed spacetimes can get away with only using SMR.
* Because the temperature is stored, the MHD source term calculation from primitive variables could be accelerated. However, now that tabulated EOSs are supported, this needs to be carefully profiled and compared with another alternate calculation based on using $T_{\mu\nu}$ directly, which already has to be calculated for Z4c.

I've added two regression tests to check that `EOSCompOSE` works (both logs and NQTs) for a simple shock tube on CPU. I'm happy to add more tests if desired.


dradice (approved):

jmstone (commented):
Looks good.  I notice a new unit test has been added.  Are these to be run as part of the CI, or as standalone tests?  Maybe there should a single "unit_tests" problem generator that runs all the unit tests?  Does the EOS for dynGRMHD use its own units class, or does it share the existing Units infrastructure used everywhere else?

jfields7:
The new unit test right now is a standalone test. I think it would be good in the future to have some way to run the different unit tests as part of the CI pipeline (or even just as a separate action which can be run manually), but I think the regression tests are likely to fail if there are problems with the features currently being tested by the existing unit tests, and then the user can run the individual tests as part of the debugging process.

The EOS for `DynGRMHD` uses its own unit infrastructure inherited from `PrimitiveSolver`, unfortunately, and it most likely does not play nicely with AthenaK's built-in unit infrastructure.

jmstone:
OK, yes let's think about adding all of the unit tests to the CI sometime in the future.  For now it is fine to keep them as stand-alone tests.  Also a shame the Units infrastructure, but understandable.  I think it is too much work to try to combine them right now.

PR #668, opened 2025-09-22
SYCL fix

dradice:
This pull requests moves some of the team barriers to avoid having issues with deadlocks on Intel hardware.


dradice:
I tested it on Intel and the unit tests ran on NVIDIA. However, I do not have access to any AMD GPUs.

PR #673, opened 2025-10-12
Add WENO-MZ reconstruction

ajdittmann:
Inspired by [Kengo's recent pull request in the Athena++ repo](https://github.com/PrincetonUniversity/athena/pull/690), I've drafted [WENO-MZ](https://onlinelibrary.wiley.com/doi/10.1002/fld.5314) as a reconstruction option here. 

Although I have only tested simple hydrodynamic problems on my laptop, it seems to be more accurate than WENO-Z. The increase in accuracy seems almost negligible for the linear wave test, but it seems to improve results for things like the Shu-Osher shock substantially. I've attached an example from that problem using PPMX, WENO-Z, and WENO-MZ with 400 cells, vs a reference run constructed using PLM and 12800 cells. WENO-Z had a slightly smaller L1 error than PPMX (~87%), and WENO-MZ had a  smaller L1 error than WENO-Z (~70%, or ~61% of PPMX's). 
shu_osher_zoom2


ajdittmann:
After further testing, it appears that WENO-MZ is unstable in GRRMHD and some highly-magnetized GRMHD contexts, occasionally leading to serious instabilities. Additionally, in some vanilla 2D MHD problems, it appears to introduce spurious high-wavenumber features. In all cases, WENO-Z and PPM remain stable, the latter typically more dissipative than the former.

PR #698, opened 2026-02-06
Add accurate GR timestep calculation option

mh-guo:
- Add 'gr_dt' boolean flag to hydro and MHD classes
- Read 'gr_dt' parameter from input file (defaults to false)
- Implement full GR wavespeed calculations when gr_dt=true
- Use conservative approximation (max speed = 1.0) when false
- Add necessary headers for metric computation


mh-guo:
A dedicated `gr_dt` flag in hydro/MHD is too special-purpose, and it only helps a few GR problems. After discussing with Jim, a better design would be a user-defined timestep function enrolled from the problem generator (along the lines of user BCs and source terms) that can replace the hydro/MHD `NewTimeStep`. That would cover the accurate GR wavespeed case without a special-purpose switch in the core classes. I'll close this PR now.

PR #708, opened 2026-02-24
Change the AMR method: tracker in z4c_amr

ruochengzhai:
Change the method `tracker` in the z4c AMR criteria. Now it checks the minimum distances from the star centers to the mesh blocks, rather than only to the 8 corners.


dradice (commented):
Please, check to make sure it can pass the test suite

dradice (approved):

PR #713, opened 2026-03-04
Cyclic zoom

mh-guo:
Implements the cyclic zoom method introduced in [Guo et al., ApJ 987, 202 (2025)](https://ui.adsabs.harvard.edu/abs/2025ApJ...987..202G/abstract) for simulating systems across a vast range of spatial and temporal scales. The method cycles between zooming in and out, using a central mask region with careful treatment of coarsened fluid variables to preserve small-scale physics.

**New module:** `src/cyclic_zoom/`
- Enabled via a new AMR criterion `method = cyclic_zoom` in the `<amr_criterion>` input block
- Cyclic AMR logic: refinement/derefinement triggers, zone advancement, and inner-boundary masking
- Cell-centered conserved variable and EMF store/restore across zoom transitions
- MHD constrained transport correction (ΔE source term) maintaining ∇·B = 0
- MPI-parallel load balancing of zoom blocks across ranks
- Full restart support: zoom state and all stored data are serialized to/from restart files

**Integration hooks:** `mesh_refinement.cpp`, `main.cpp`

**New regression tests:** `tst/test_suite/zoom/`

**New input files:** `inputs/zoom/`


jmstone (commented):
Certainly a major feature upgrade, thanks Minghao.  A few suggestions:
(1) Cyclic zoom needs documentation
(2) The physics modules with which cyclic zoom works should be documented.  It would be good to add checks in the constructor that terminate execution with a warning if cyclic zoom is used with physics modules that are not implemented
(3) Overall looks good, but as we use it for more applications we may decide to make changes.  You should document anything you think could be areas for improvement.

mh-guo:
Thank you Jim.

(1) Added a Cyclic Zoom page in athenak-docs: [#12](https://github.com/IAS-Astrophysics/athenak-docs/pull/12)

(2) That page documents which physics is supported. The constructor now exits if cyclic zoom is used with unimplemented modules or a 1D/2D mesh.

(3) Known limitations / possible future changes are listed on the same page.

jmstone:
Thanks for the updates.  I also asked David Velasco to review and all looks good to him too, so here goes with the merge!

PR #725, opened 2026-04-02
feat: add the skeleton of the Chemistry module

bcaddy:
## Summary

This adds a skeleton version of the chemistry module. That skeleton is currently limited to adding extra passive scalars, writing a dummy value to them, and naming them in output files. The chemistry module is hooked into the rest of AthenaK by instantiating the `Chemistry` object in either the hydro or MHD integrators.

This also adds a new input file in the new `inputs/chemistry` directory. It's just a copy of the sod shock tube with chemistry turned on for testing and it will be removed when a more appropriate chemistry input file is created.

Since the chemistry module isn't complete yet it will print a warning message if it's used.

## Other changes:
- added `__pycache__` and `tst/test_log.txt` to `.gitignore`

## Requests for Reviewers

1. I'm especially interested in feedback on structure and if I'm doing things the "AthenaK way" to maintain consistent style. 
2. ~~I'm updating `nscalars` implicitly within the chemistry constructor. I don't love this since it means that the constructor has side effects but the alternative was to require that `nscalars` be updated in a separate call which also seemed potentially error prone to me.~~ This is no longer true now that `Chemistry` is instantiated in `AddPhysics`
3. Would it be better to instantiate chemistry within `MeshBlockPack::AddPhysics`? That looked like it would make modifying the number of passive scalars more complicated.
    - This has been done

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ tst/test_log.txt
 *~
 *.swp
 Lorene
+
+__pycache__
+tst/test_log.txt
--- a/inputs/chemistry/sod_with_chemistry.athinput
+++ b/inputs/chemistry/sod_with_chemistry.athinput
@@ -45,6 +45,7 @@ eos         = ideal    # EOS type
 reconstruct = plm      # spatial reconstruction method
 rsolver     = llf      # Riemann-solver to be used
 gamma       = 1.4      # gamma = C_p/C_v
+nscalars    = 2        # number of passive scalars in hydro
 
 <chemistry>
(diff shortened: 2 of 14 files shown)


bcaddy:
I added a small bit of code to change the names of the passive scalars reserved for chemistry in the output file. Instead of "s_xy" they're now "s_xy_species_name". Currently I'm just using stand-ins for the names but in the future this should make it easier to differentiate between the different passive scalars.

bcaddy:
Moved to [here](https://github.com/bcaddy/athenak/pull/3)

PR #726, opened 2026-04-06
pgen for Kadath FUKa_v2 BNS initial data

alanlam1002:
Add pgen for binary neutron star from initial data solver Kadath FUKa_v2. Both MPI and OpenMP parallelisation are supported.

### DIFF
--- a/src/pgen/kadath_bns.cpp
+++ b/src/pgen/kadath_bns.cpp
@@ -38,6 +38,10 @@
 #include "parameter_input.hpp"
 #include "z4c/z4c.hpp"
 #include "z4c/z4c_amr.hpp"
+#include "utils/tov/tov_utils.hpp"
+#include "utils/tov/tov_polytrope.hpp"
+#include "utils/tov/tov_piecewise_poly.hpp"
+#include "utils/tov/tov_tabulated.hpp"
 
 // ---------------------------------------------------------------------------
 // Output index enumeration for KadathExportBNS.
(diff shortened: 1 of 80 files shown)


src/pgen/kadath_bns.cpp
```diff
@@ -0,0 +1,579 @@
...
+    } else {
+      if (use_cold_table) {
+        using eos_t = Kadath::Margherita::Cold_Table;
+        host_w0(m, IDN, k, j, i) = EOS<eos_t, DENSITY>::get(h_enth);
+        host_w0(m, IPR, k, j, i) = EOS<eos_t, PRESSURE>::get(h_enth);
+      } else if (use_cold_pwpoly) {
+        using eos_t = Kadath::Margherita::Cold_PWPoly;
+        host_w0(m, IDN, k, j, i) = EOS<eos_t, DENSITY>::get(h_enth);
+        host_w0(m, IPR, k, j, i) = EOS<eos_t, PRESSURE>::get(h_enth);
+      }
```
jfields7, line 380: I am assuming that FUKA does some sort of rescaling of the baryon mass to ensure that $h = 1$ at the surface. Does it scale it back to the original value when you read in things like the density and pressure? If not, it will be safer to do what is done in the Lorene and Elliptica pgens and get the total energy density $\rho(1 + \epsilon)$, which is invariant under rescaling, and then recompute the density and pressure from that energy.

alanlam1002: I think FUKA only stores the specific enthalpy $h$ in the initial data, and obtains the energy density, rest-mass density, and pressure directly from $h$.

jfields7: Okay, but the value of specific enthalpy depends on what you choose for the baryon mass. If FUKA is rescaling it, the rest-mass density it returns will be in the wrong units.

alanlam1002: As far as I know, FUKa does not rescale the specific enthalpy. The mass unit remains fixed throughout the iteration. FUKa calls these the hydro-rescaling stages, in which it rescales $$h$$ to fix the baryonic/rest-mass during each iteration. However, FUKa does not rescale the mapping between $$h$$ and density $$\rho$$.

I have also prepared the instructions for running the kadath_bns pgen, which can be found in
https://github.com/alanlam1002/athenak-tutorial-kadath

PR #730, opened 2026-04-20
Fix memory leak in particles MPI communication (#716)

Miriam-R-coder:
Fixes #716. Moved mpi_ituple creation to constructor/destructor to prevent memory leak.


jmstone:
This was fixed with #767.  Your fix of declaring/destroying the data only once in the constructor/destructor is more efficient, and maybe we'll switch to that in the upcoming re-design of the particle modules.

PR #734, opened 2026-04-27
Fix derived output variable allocation after AMR changes

### LINKED ISSUE #646: Derived variables incompatible with AMR
Derived variables don't seem to be compatible with AMR and can cause the code to segfault during output after refinement. The culprit seems to be lines like the following inside `ComputeDerivedVariable` inside `derived_variables.cpp`:
```c++
if (name.compare("...") == 0) {
  if (derived_var.extent(4) <= 1)
    Kokkos::realloc(derived_var, nmb, n_dv, n3, n2, n1);
  ...
}
```

The simplest solution is to disable the check `if (derived_var.extent(4) <= 1)`, but this means memory for derived variables will be reallocated during every output. Is there a better solution for this?

mh-guo:
Reallocate derived output buffers to the current meshblock pack shape before computing each derived variable. This avoids stale buffer extents after AMR changes `nmb_thispack`, which could trigger out-of-bounds GPU writes during output.

### DIFF
--- a/src/outputs/derived_variables.cpp
+++ b/src/outputs/derived_variables.cpp
@@ -94,8 +94,7 @@ void BaseTypeOutput::ComputeDerivedVariable(std::string name, Mesh *pm) {
 
   // temperature = pressure / density
   if (name.compare("temperature") == 0) {
-    if (derived_var.extent(4) <= 1)
-      Kokkos::realloc(derived_var, nmb, n_dv, n3, n2, n1);
+    Kokkos::realloc(derived_var, nmb, n_dv, n3, n2, n1);
     auto dv = derived_var;
     auto &w0_ = (name.compare("hydro_wz") == 0)?
       pm->pmb_pack->phydro->w0 : pm->pmb_pack->pmhd->w0;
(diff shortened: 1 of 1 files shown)


ruochengzhai:
Hi @mh-guo, I wonder if it would save some time to allocate the derived variables using `max_nmb_per_rank`, as with the mhd variables. In this way, they'll not be reallocated every time they are used. However, one drawback of this method is that I need two nmb variables for allocation and calculation. I have implemented this in the branch [ls-project](https://github.com/IAS-Astrophysics/athenak/tree/ls-project).

jmstone:
I agree that re-allocating the derived variables every time an output is made is not the best solution.  Could you update this PR so that they are only allocated once per run, as in Ruo Cheng's comment?

mh-guo:
Thank you @ruochengzhai  for the suggestions! I have updated it. In addition, there are a few extra bugs fixed, including moments buffer dims (these blocks are currently unregistered, so unused), a missing mhd_k_jxb index increment, and an mhd_dynamo_ks index typo.

PR #739, opened 2026-05-08
Fix deep-AMR face-centered divB preservation

dfielding14:
## Summary

This PR builds directly on the work started by Hengrui Zhu in #732.

PR #732 identified an important problem in AMR face-centered magnetic-field prolongation: a coarse-neighbor boundary path could overwrite active faces that should be owned by a same-level or finer neighbor. That is a real bug, and fixing that overwrite path is the right starting point.

While testing that fix, we added a stricter moving-AMR `div(B)` regression. That test shows that the #732 fix is necessary but not quite sufficient for deeper 2D/3D AMR. The remaining problem is that several AMR rebuild paths can still leave the face-centered magnetic field internally inconsistent after refinement/derefinement and boundary exchange.

This PR keeps the same basic idea as #732, but applies it more generally:

- active-domain face-centered fields should have a clear owner;
- coarse-neighbor prolongation should not write active faces owned by same-level or finer neighbors;
- restricted coarse face fields should be current before AMR packing;
- after an AMR rebuild, internal fine faces that depend on finalized exterior faces should be repaired using the existing Toth-Roe construction;
- MHD primitives should then be refreshed from the repaired face fields.

This is not a divergence-cleaning or projection step. The goal is just to make the existing CT/Toth-Roe AMR construction happen in the right ownership/order so that the discrete face-centered `div(B)` constraint remains at roundoff.

## What Was Going Wrong

The #732 fix handles one important overwrite case in boundary prolongation. The stricter test added here exposed two additional issues.

First, the ownership rule was still too broad. Some edge/corner AMR boundary paths contain a mixture of faces: some are truly owned by the coarse-neighbor prolongation path, while others are owned by same-level or finer neighbors. A range-level clip can miss those mixed cases. This PR replaces that with exact active-face ownership checks for the face-centered prolongation writes.

Second, AMR mesh rebuilds can compute internal fine faces before all exterior faces have reached their final values. In 2D and 3D, the Toth-Roe internal face construction depends on the surrounding exterior face fluxes. If those exterior faces are later changed by boundary exchange or prolongation, the previously computed internal faces are no longer the divergence-preserving ones for the final exterior data.

The repair pass added here addresses only that specific issue: after boundary communication has finalized exterior faces, it recomputes the affected internal face-centered fields using the existing Toth-Roe construction. It leaves exterior block faces alone.

## New Regression Test

This PR adds a new `divb_amr` regression test specifically designed to stress AMR face-centered magnetic-field preservation.

The test initializes the magnetic field from a discrete staggered vector potential, then takes the discrete curl to fill the face-centered magnetic field. That means the initial field is divergence-free in the same finite-volume sense used by CT. A uniform guide field is added, and the gas is given a mild velocity so the field evolves while the AMR hierarchy changes.

The AMR pattern is deliberately artificial. Instead of refining based on the fluid state, the test uses a user refinement callback with several moving periodic refinement boxes. Those boxes move across the domain and force repeated refine/derefine cycles. This creates many coarse/fine faces, edges, and corners, including the deeper multi-level cases where the original problem shows up.

The history diagnostic directly measures the discrete divergence from the evolved face-centered fields on active cells:

- `max_ndiv`: maximum normalized `|div(B)|`;
- `L1`: volume-weighted mean normalized `|div(B)|`;
- `L2`: volume-weighted RMS normalized `|div(B)|`;
- `ncell`: used to confirm that the test actually refined above the root grid.

The regression runs:

- 1D moving AMR;
- 2D moving AMR;
- 3D moving AMR with 1 through 5 physical refinement levels.

The acceptance thresholds are intentionally strict:

- `max_ndiv < 2e-11`;
- `L1 < 2e-12`;
- `L2 < 5e-12`.

This is meant to catch exactly the kind of failure that is easy to miss if only one or two AMR levels are tested.

## Baseline Results

All cases below use the same `divb_amr` test and the same thresholds.

| Case | upstream/main | PR #732 branch | this PR |
|---|---:|---:|---:|
| 1D | pass | pass | pass |
| 2D | fail | fail | pass |
| 3D, 1 physical AMR level | pass | pass | pass |
| 3D, 2 physical AMR levels | fail | fail | pass |
| 3D, 3 physical AMR levels | fail | fail | pass |
| 3D, 4 physical AMR levels | fail | fail | pass |
| 3D, 5 physical AMR levels | fail | fail | pass |

Representative passing values on this branch are roundoff-level:

| Case | max_ndiv | L1 | L2 |
|---|---:|---:|---:|
| 1D | 0 | 0 | 0 |
| 2D | 2.21e-15 | 4.14e-17 | 5.30e-17 |
| 3D L1 | 3.20e-15 | 4.81e-17 | 6.28e-17 |
| 3D L2 | 3.58e-15 | 4.95e-17 | 6.49e-17 |
| 3D L3 | 3.55e-15 | 4.93e-17 | 6.48e-17 |
| 3D L4 | 3.90e-15 | 5.04e-17 | 6.65e-17 |
| 3D L5 | 3.64e-15 | 5.05e-17 | 6.66e-17 |

## Validation

On this branch:

- `git diff --check`: pass
- `python3 -m py_compile tst/scripts/mhd/mhd_divb_amr.py`: pass
- MPI build of `athena` with `-DAthena_ENABLE_MPI=ON`: pass
- 8-rank MPI `divb_amr` sweep: pass for 1D, 2D, and 3D with 1-5 physical AMR levels

## Notes for Reviewers

The patch is intentionally narrow. It only touches the generic MHD AMR face-centered field handling and adds the regression needed to reproduce the issue. It does not change file formats, input-file interfaces, or physics options.

The main conceptual change is that AMR face-centered fields now follow a stricter ownership/order rule: exterior faces are finalized first, then affected internal fine faces are recomputed from those final exterior values using the existing divergence-preserving construction.


dfielding14:
Thanks again @HengruiZhu99 for starting this in #732. I opened #739 as a follow-up that keeps the same basic direction, but extends the fix to the deeper AMR cases exposed by a stricter moving-AMR `div(B)` test.

The short version is:

- #732 fixes a real coarse-neighbor prolongation overwrite path.
- The new `divb_amr` regression shows that this is necessary but not sufficient for 2D/deep 3D AMR.
- The remaining issue is that AMR rebuilds can leave internal fine faces inconsistent with exterior faces after boundary exchange/prolongation.
- #739 adds exact active-face ownership checks, refreshes restricted FC data before AMR packing, and repairs only topology-changed blocks after exterior faces are finalized.
- The repair is not divergence cleaning or projection; it reuses the existing Toth-Roe construction.

The new regression tests 1D, 2D, and 3D moving AMR, including 3D sweeps with 1-5 physical refinement levels. With the same thresholds, `main` and #732 fail the 2D/deep-3D cases, while #739 passes at roundoff.

CI is now green on #739:
https://github.com/IAS-Astrophysics/athenak/pull/739

Tagging @dradice @jmstone @jfields7 since you were already requested on #732. I am happy for this to be reviewed either as a replacement/follow-up PR, or for the single commit in #739 to be cherry-picked onto the existing `fix_amr` branch if that is easier.

dfielding14:
As an additional test I ran a driven 3D MHD turbulence case with nested AMR. The figure below shows a final x3 slice through the magnetic field, magnetic pressure, beta, current magnitude, normalized discrete `div(B)`, and AMR level.

The main point is the bottom-middle panel: even with turbulent structure crossing several refinement boundaries, the normalized discrete `|∇·B| Δx / |B|` remains at roundoff, while the magnetic-field components do not show visible AMR imprinting at the refinement interfaces.

image

HengruiZhu99:
I will test this on my circumbinary disk and post update here. Thanks @dfielding14 !

jmstone:
Thanks Drummond, finally getting around to merging this and looks great.

You've added a regression test in the /tst/scripts directory, but a recently that has been deprecated and we've switched to only using pytest, with scripts in the /tst/test_suite subdirectories.  Could you move the script there?  You'll have to decide whether it should run on the cpu or GPU and give the script the appropriate name.  How long does the test take?  Each test should be no more than 30s to keep the total test time manageable.  Perhaps it would be OK to only run it in 2D/3D (certainly divB shouldn't be a problem in 1D).

dfielding14:
Thanks Jim. Done—I moved the regression into `tst/test_suite/nr/test_nr_divb_amr_mpicpu.py`. I chose MPI CPU because this exercises generic MHD AMR boundary exchange rather than GPU-specific behavior, while also providing cross-rank coverage.

I dropped the 1D case and retained two focused cases: 2D and 3D with two physical refinement levels, which is the minimum 3D case that exposes the bug. On my local eight-rank MPI-CPU run, they take 0.49 s and 1.32 s, respectively, or 1.84 s together.

Both remain effective regressions: the fixed code stays at roundoff, while an isolated pre-fix control fails by many orders of magnitude. I also moved the inputs into `tst/inputs`, removed the deprecated script and unused 1D input, and the style checks pass.

jmstone:
Beautiful. I'm impressed by how quick the tests run, and you've done all the right things, e.g. moving the inputs to /tst/inputs.

However, now there is a conflict I am afraid, involving code inside an ifdef on ATHENAK_DEBUG_FC_AMR_OWNERSHIP.  Can we perhaps remove this debugging code everywhere, and hopefully that would also get rid of the conflict?

dfielding14:
Ah yes, of course—sorry about that. I’ve removed all of the debugging code everywhere, and the conflict is now resolved.

jmstone:
Perfect, thanks Drummond!

PR #741, opened 2026-05-19
Add basic implementation of chemistry with H2 network and Forward Euler ODE solver

bcaddy:
_note: this currently also includes the changes from PR #725. Once that PR is merged I will rebase this PR so the diff is clearer. Or this PR can be merged directly and PR #725 closed._

# Summary

This PR adds a new H2 chemistry network, a general explicit Forward Euler ODE solver class, and uses both of them in the `Chemistry` class to add fully functional chemistry to AthenaK. 

## Forward Euler Solver

This is a port of the Forward Euler solver used in Athena++'s Chemistry module. I've changed it a little bit so that it works for general systems of ODEs and can be used by any module in AthenaK. It takes the system of ODEs to solve as a template argument and requires that the ODE system class have the following members: 

- A `neqs` variable that specifies the number of equations. That should be the number of species plus 1 for the internal energy
- `y` and `f` `RegisterArray` variables to hold the current state and the result of evaluating the equations respectively.
- An `evaluate_function` method that computes `f` from `y`

It also accepts the following runtime arguments: 

| Option               | Type   | Default | Description                                                                                      |
| -------------------- | ------ | ------- | ------------------------------------------------------------------------------------------------ |
| fe_cfl               | Real   | 0.1     | cfl number for subcycling, lowering it has virtually no impact on the solution (for H2 at least) |
| fe_n_subcycle_max    | int    | 1e5     | maximum number of substeps                                                                       |
| fe_yfloor            | Real   | 1e-3    | The minimum y value of the ODEs allowed                                                          |

Since the user might want to use the same ODE solver for different modules with different settings these settings must be defined under the proper module block. An example of how to choose the correct templated version of the ODE solver and passing it the proper arguments can be found in the `Chemistry::UpdateChemistryTask()` function within `src/chemistry/chemistry.cpp`.

Note: In the Athena++ implementation the `yfloor` is only applied to the chemistry scalars. Since this is a general solver, not just for chemistry, there isn't a good way to do that so the floor applies to the internal energy as well. Let me know if this is an issue and we can figure out a solution.

## H2 Network

This is an implementation of the H2 network from [Gong et al. 2023](https://ui.adsabs.harvard.edu/abs/2023ApJS..268...42G/abstract) and is a pretty direct port from Athena++. The biggest change is combining the functions that update internal energy and the chemistry scalars into one as the ODE solver requires. Options:

| Option                 | Type   | Allowed Values | Default       | Description                                                                        |
| ---------------------- | ------ | -------------- | ------------- | ---------------------------------------------------------------------------------- |
| h2_constant_cv         | bool   | true, false    | false         | Whether or not C_v should be constant. If true then C_v is set to 1.65*K_boltzmann. Used to make the solution analytically solvable for testing. |
| h2_isothermal          | bool   | true, false    | false         | If the H2 network should use an isothermal EOS, i.e. set the change in internal energy to zero |

## Chemistry 

This also converts the `Chemistry` class from just a skeleton to actually performing the chemistry update. Right now it uses the H2 network and Forward Euler ODE solver to evolve chemistry. It performs the chemistry solve on all cells, both ghost and real cells, to avoid having to re-communicate the ghost cells. Once we have a more realistic network and ODE solver we can profile it and figure out if that really is the best option. Options:

| Option                 | Type   | Allowed Values | Default       | Description                                                                        |
| ---------------------- | ------ | -------------- | ------------- | ---------------------------------------------------------------------------------- |
| network                | string | H2      | H2            | The network to use                                                                 |
| ode_solver             | string | forward_euler  | forward_euler | The ODE solver to use                                                              |

### Tests

This adds several chemistry tests that run on CPU, GPU, and MPI-CPU. The `H2_uniform` test evolves a uniform field with no hydrodynamic evolution and the `H2_advecting` test advects a Gaussian in the chemical state across the domain. Both are compared to analytical results and the advection test also performs a convergence test.

## Other

- Added the `RegisterArray` class to `src/utils/register_array.hpp` that can be used to provide in-register arrays on GPUs that have View syntax, namely `size` and access with `operator()`

## Proposed Wiki Changes

Adding this block to `The-Input-File.md` 

> * `<chemistry>` (optional) : parameters of chemistry
>   * `network`: (optional) The chemistry network to use. Options include: `H2`, and `GOW17`(upcoming). These networks often have their own runtime parameters which are detailed in the [Chemistry](Chemistry) page.
>   * `ode_solver`: (optional) The ODE solver to use. See the [ODE Solver](ODE-Solver) page for which solvers are available and any runtime parameters they may have.

Along with new "Chemistry" and "ODE Solvers" pages that will be linked in the `_sidebar.md`

[Chemistry.md](https://github.com/user-attachments/files/28474669/Chemistry.md)

[ODE-Solvers.md](https://github.com/user-attachments/files/28319405/ODE-Solvers.md)


bcaddy:
I've added the proposed changes to the wiki as well. The chemistry page is based on the chemistry page in Athena++.

src/chemistry/network/H2.hpp
```diff
@@ -0,0 +1,130 @@
...
+  bool const_cv;
+};
+
+class H2Network {
+ public:
+  KOKKOS_FUNCTION H2Network(H2Settings const settings, Real const density,
+                            Real const density_cgs, Real const mu,
+                            Real const hydrogen_mass_cgs,
+                            Real const units_time_cgs,
+                            Real const units_energy_density_cgs)
```
munan, line 43: Wouldn't it be better to just parse the unit class?

src/chemistry/network/H2.hpp
```diff
@@ -0,0 +1,130 @@
...
+   */
+  static H2Settings GetSettings(ParameterInput* pin) {
+    return H2Settings{pin->GetOrAddBoolean("problem", "h2_constant_cv", false)};
+  }
+
+  /*!
+   * \brief Updates `f` using the values in `y`
+   */
+  KOKKOS_FUNCTION
+  void evaluate_function() {
```
munan: I may call it "evaluate_rhs" or something, where rhs stands for the right-hand-side of ODE. "function" seems a bit vague.

src/ode_solvers/forward_euler.hpp
```diff
@@ -0,0 +1,147 @@
...
+#include "athena.hpp"
+
+namespace ode_solvers {
+
+struct FESettings {
+  /// The max number of subcycles the Forwared Euler solver should execute
+  unsigned int fe_n_subcycle_max;
+  /// CFL number for the subcycles
+  Real fe_cfl;
+  /// The minimum y value of the ODEs allowed
```
munan: This is not the floor for chemical abundances from the ODE solver, this is just for calculating the chemical timescale for sub-cycling.

src/outputs/basetype_output.cpp
```diff
@@ -294,6 +295,14 @@ BaseTypeOutput::BaseTypeOutput(ParameterInput *pin, Mesh *pm, OutputParameters o
         std::string vname;
         vname.assign("s_");
         vname.append(number);
+
+        if (pin->DoesBlockExist("chemistry")) {
+          if(n>=pm->pmb_pack->pchemistry->get_chemistry_scalars_first_idx()
+             && n<=pm->pmb_pack->pchemistry->get_chemistry_scalars_last_idx()) {
+              vname.append("_"+pm->pmb_pack->pchemistry->GetSpeciesNames(n));
```
munan, line 310: In Athena++ output, s_[species_name] is the conserved variable (relative abundances x density), and r_[species_name] is the primitive variable (relative abundance). Is this also true here?

src/ode_solvers/forward_euler.hpp
```diff
@@ -0,0 +1,147 @@
...
+      {
+        // calculate chemistry timescale
+        dt_subcycle = Kokkos::reduction_identity<Real>::min();
+        for (int s_idx = 0; s_idx < ode_system.neqs; s_idx++) {
+          // put floor in species abundance
+          Real const yf = Kokkos::max(ode_system.y(s_idx), fe_yfloor);
+
+          // Compute the value to reduce
+          // NOLINTNEXTLINE(build/include_what_you_use)
+          dt_subcycle = Kokkos::min(
```
munan, line 116: Does this account for the cooling time?

src/ode_solvers/forward_euler.hpp
```diff
@@ -0,0 +1,147 @@
...
+      ode_system.evaluate_function();
+
+      // Compute the chemistry time scale
+      Real dt_subcycle;
+      {
+        // calculate chemistry timescale
+        dt_subcycle = Kokkos::reduction_identity<Real>::min();
+        for (int s_idx = 0; s_idx < ode_system.neqs; s_idx++) {
+          // put floor in species abundance
+          Real const yf = Kokkos::max(ode_system.y(s_idx), fe_yfloor);
```
munan: Since you are looping over s_idx from 0, it looks like you are putting fe_yfloor also on the internal energy?

src/ode_solvers/forward_euler.hpp
munan: Looks like you are treating internal energy the same as chemical abundances, right?

src/chemistry/network/H2.hpp
munan: Maybe it is OK if you really include both the energy equation and the chemical reaction equations

src/ode_solvers/forward_euler.hpp
```diff
@@ -0,0 +1,147 @@
...
+  void SolveODE() {
+    // ------ Solve the ODEs ------
+    unsigned int icount = 0;
+    Real t_now = t_start;
+    Real t_end = t_start + dt;
+    while (t_now < t_end) {
+      // Evaluate the ODEs
+      ode_system.evaluate_function();
+
+      // Compute the chemistry time scale
```
munan: I may call this "chemistry and heating/cooling timescales", since you are also computing energy equation

src/chemistry/network/H2.hpp
bcaddy: Yeah, but the units class lives in the host (i.e. CPU) memory space and isn't accessible on the GPU and this constructor is initialized on the GPU. I tried initializing the chemistry network and ODE solver networks on the host then copying them to the GPU and it was much more complex and confusing for little gain.

src/chemistry/network/H2.hpp
bcaddy: I picked this since it's what the Kokkos Kernels ODE solvers require but I agree that it's not as clear as I would like it to be. Once we sort out the ODE solvers I might come back and rename things.

src/ode_solvers/forward_euler.hpp
bcaddy: Fixed in cce20af

src/outputs/basetype_output.cpp
bcaddy: It looks like it's the opposite in AthenaK, `r_[species name]` for conserved variables and `s_[species name]` for primitive variables, no clue why. I added this block to the conserved outputs in cafc60a so that both have the species name.

src/ode_solvers/forward_euler.hpp
bcaddy: Yeah, I'm including internal energy in the loop.

src/ode_solvers/forward_euler.hpp
bcaddy: Yeah, I tried to exclude it but trying to keep the ODE solver general made that ugly pretty quick. If it's required to not floor internal energy I can figure out a way but if not then this way is a lot less complex.

src/ode_solvers/forward_euler.hpp
bcaddy: Fixed in 93804e2. I made the comments generic to any ODE since this is supposed to be a general solver

src/chemistry/network/H2.hpp
munan: I see. This is very problem specific though. Another idea is to have basic unit inputs (length, mass and time in Units class) and compute the other units based on those (similar to the void `Units::SetUnitsConstants()` function), or use the basic unit inputs to construct a unit class inside the chemical network class. This would be a more general approach. You can see if this makes sense when you implement the GOW17 network which uses more unit conversions than the simple H2 network, and doing a lot of input unit parameters may be tedious.

src/chemistry/network/H2.hpp
munan: If you are seperating reaction and energy source terms then you can name them seperately and put them together under evaluate_function()

src/outputs/basetype_output.cpp
munan: Ah this is confusing. Can you change it to be the same as Athena++? Or it is a desgin choice made in the passive scaler output that's already done?

src/ode_solvers/forward_euler.hpp
munan: So the advantage of putting a species floor is that the species abundances returned by the ODE solver can sometimes be negative. I think the Reiman solver can also sometimes give negative internal energy, but it's less common. If you are separating the reaction and energy equations, would it be easy to implement different floors for species and internal energy? I think that may make more sense for the physics we are solving. For example, one may want to have yfloor similar to the smallest meaningful chemical abundance for species, and the smallest meaningful temperature for energy.

When you get to test GOW17 network, I would think you want to use a smaller yfloor for chemical species. Perhaps something on the order of 1e-12 since the species abundances can be very small and still meaningful. Actually, could you test using the default of a small number like 1e-12 for the H2 regression tests? If it works, I may put a smaller number as the default, since a large yfloor may result in the solver taking too large a timestep and not converge (probably not a problem in the H2 tests but can be for the GOW17 network).

src/chemistry/network/H2.hpp
bcaddy: The first argument is a `struct` that can contain anything and is used right now mostly to get parameters from the input file, worst case we add it there. I'd like to keep as much of the the units stuff in the `Units` class as possible but I'll revisit this as I'm implementing the GOW17 network.

src/chemistry/network/H2.hpp
bcaddy: That's what I'm planning on doing

src/outputs/basetype_output.cpp
bcaddy: It is a design choice made in the passive scaler output that's already done and predates my involvement. Once we move to using sparse fields instead of passive scalars I can use more logical naming.

src/ode_solvers/forward_euler.hpp
bcaddy: A general ODE solver won't know about the difference between species and internal energy, it's just going to treat them both as ODEs. I could do something like "all ODEs with indices above X get different or no floor" reasonably easily and just put internal energy at the end of the array instead of the beginning (I'm moving it there anyway to make the creation/destruction functions simpler.)

munan: I think if you use a default small yfloor (like 1e-12) it should be OK for both species and internal energy in most cases. Could you give it a try for the H2 test? I am just concerned that a default 1e-3 is too large.

inputs/chemistry/H2_uniform.athinput
```diff
@@ -0,0 +1,78 @@
...
+gamma           = 1.666666666666667 # gamma = C_p/C_v
+nscalars        = 0                 # number of passive scalars in hydro
+iso_sound_speed = 0.6               # isothermal sound speed
+
+<units> 
+# ISM units
+length_cgs  = 3.0856775809623245e+18  # length is 1 pc
+mass_cgs    = 6.195900228622575e+31   # number density is 1 cm^-3
+time_cgs    = 3.15576e+13             # time is 1 Myr
+mu          = 1.4                     # mean molecular weight
```
jeonggyukim: This variable should be named muH or mu_hyd (rho = muH*n_H*m_H), the mean molecular weight per H nucleon. Usually, mu is reserved for mean molecular weight per particle.

bcaddy: This is the mean molecular weight per particle. It predates chemistry. Does chemistry use $\mu_H$ and not $\mu$?

src/chemistry/thermo/thermo.hpp
```diff
@@ -0,0 +1,43 @@
...
+  //! \brief This computes the specific heat (C_v) of a cold gas, i.e. assuming
+  //! that H2 rotational and vibrational levels not excited.
+  //!
+  //! xH2, xe = nH2 or ne / nH
+  //! xHe_total = xHeI + xHeII = 0.1 for solar value.
+  //! Return: specific heat per H atom.
+  KOKKOS_FUNCTION static Real CvCold(const Real xH2, const Real xHe_total,
+                                     const Real xe) {
+    const Real xH20 = (xH2 > 0.5) ? 0.5 : xH2;
+    return 1.5 * units::Units::k_boltzmann_cgs *
```
jeonggyukim: So 1.5 (= 1 / (gamma -1)) is hardcoded here. Would it be better to have `gamma` as an input parameter for chemistry (we can inherit from hydro's gamma)? @munan what do you think?

inputs/chemistry/H2_uniform.athinput
jeonggyukim: Judging from its numerical value 1.4 and where it is used (converting rho to nH)

`n_H(density * density_cgs / (mu * hydrogen_mass_cgs))`

it should be named muH. The mean molecular weight per particle is related to muH as:
`mu = muH/(1. - 2. * xH20) + xH20 + xHe_total + xe)`
It appears mu is folded in `CvCold`.

src/chemistry/thermo/thermo.hpp
bcaddy: Yes, I can make that change so it inherits from the EOS

inputs/chemistry/H2_uniform.athinput
bcaddy: I can add $\mu_H$ to the units class pretty easily, though coupling $\mu$ and $\mu_H$ the way you describe would be more challenging and really only effect other modules so I'll probably leave them both as parameters you can set at runtime for now. Does that work?

src/chemistry/thermo/thermo.hpp
bcaddy: fixed

inputs/chemistry/H2_uniform.athinput
bcaddy: I ended up adding `mu_H` to the chemistry module, it's easy enough to change later if needed.

jeonggyukim: Yes, this would work!

src/chemistry/network/H2.hpp
jeonggyukim: neqs = 3 but two species names, which makes third element empty. Should we make this `nesq - 1`?

bcaddy: the number of equations (`neqs`) includes the internal energy since it's part of the ODE system. `Chemistry::GetSpeciesNames` handles using `species_name` for outputs, it's not used anywhere else.

jeonggyukim: Right, but isn't the number of species names we would need for chemistry output `neqs-1`?
(internal energy is part of the hydro variables)

bcaddy: Oh, I see what you mean now. Yeah, the third element is empty, I'll fix it

bcaddy:
The documentation for this PR is now staged in [PR #5](https://github.com/IAS-Astrophysics/athenak-docs/pull/5) in the AthenaK docs repo

bcaddy:
Moved to [here](https://github.com/bcaddy/athenak/pull/2)

PR #743, opened 2026-06-01
Apparent horizon finder and smooth excision

### LINKED ISSUE #634: Excision atmosphere for DynGRMHD
The existing excision treatment for `DynGRMHD` currently uses the same `coord/dexcise` and `coord/pexcise` variables as the rest of the code. This works well enough for something like an ideal gas where the pressure is easy to calculate, but it's inconsistent with `PrimitiveSolver` using the temperature (`mhd/tfloor`) for the floor. It would be better to have an optional `coord/texcise` variable as well as a consistent treatment for passive scalars in excised regions.

spacetimecurv:
This PR adds the fast-flow apparent horizon finder and some other utilities to AthenaK. 

## Apparent horizon finder

The horizon finder is based on https://arxiv.org/pdf/gr-qc/9707050 and is widely used in various NR codes. I have re-written the version in GR-Athena++ for AthenaK and made it such that the heavy computation can be done entirely on the GPU. The biggest bottleneck is the interpolation of the metric fields onto the trial surface inside the fast-flow algorithm. I overcame this problem by adding an inline interpolator which is a direct mirror of the Lagrange interpolator class, but which can be used inside Kokkos kernels. This results in significantly improved performance. For the computation of the spherical harmonics I also added functions for the computation of the first and second derivatives. The horizon finder is interfaced with the Compact Object Trackers. Inside the ```CompactObjectTracker``` class there is a parameter for the ```mass``` which provides an initial guess for the radius of the trial surface based on the mass-radius connection in isotropic coordinates. Alternatively, one can specify an initial radius by hand and once the surface has been found, the last found radius is used as initial guess. The user has the choice between three different flow functions used in the computation of the surface integrals. Some early benchmark tests are shown below (input files are also provided).

Single puncture:
hor_sum

Spinning puncture:
hor_sum_spinning

BBH (unequal-/equal-mass):
bbh_horsum

Note: in the latest version of the horizon finder the spin for the single non-spinning puncture, which should be close to machine precision, will be orders of magnitude larger. The reason is, that the surface integral computation is now done inside a Kokkos kernel and we cannot guarantee direct cancellation anymore.

## Smooth excision

For runs in dynamical GRMHD, where a BH is present, we also include the option for smooth excision which is based on the following prescription, where we add some damping source terms to the RHS of the matter equations:

$$
\partial_t \tilde{D} = \ldots - w (\tilde{D} - \tilde{\rho}_\mathrm{ex}) / t_\mathrm{damp},
$$

$$
\partial_t \tilde{S}_i = \ldots - w \tilde S_i / t_\mathrm{damp},
$$

$$
\partial_t \tilde{\tau} = \ldots - w (\tilde{\tau} - \tilde{\tau}_\mathrm{ex}) / t_\mathrm{damp},
$$

$$
\partial_t (\tilde{D} Y_e) = \ldots - w (\tilde{D} - \tilde{\rho}_\mathrm{ex}) Y_e / t_\mathrm{damp}.
$$

As a test case we embedded a single puncture in a constant background of a certain density and pressure. Matter is damped over the specified timescale that should be larger than the time step to ensure stability. Some weird coupling between MHD+Z4c always lead to negative $$\chi$$ close to the puncture, which was circumvented by adding an artificial floor on $$\chi$$ after the RK update. This is not ideal, but solves the issue short-term.

### 1. Non-smooth excision
Evolution:
https://github.com/user-attachments/assets/74475207-2da1-406c-8e65-c2e4173cdf45

Lapse evolution:
https://github.com/user-attachments/assets/641897ea-d174-442a-be3c-635726f6474e

### 2. Smooth excision
Evolution:
https://github.com/user-attachments/assets/5daceba3-20ee-4f21-afb7-cfcd171811d4

### 3. Comparison:
https://github.com/user-attachments/assets/1a3eb69b-73a3-4d42-abf3-bc9baba1c483

The simulations above were created with an ideal gas EOS. Some of the spikes at early times are likely coming from the propagation of a gauge wave (see lapse evolution), that results in a "bounce" of the lapse. In the smooth case we can see a nice profile for the density at late times, while the profiles for the pressure and temperature are likely dominated by the EOS prescription and problem setup.

## Elliptica/Lorene pgen
Some minimal changes/fixes have been implemented in the Elliptica and Lorene pgen, i.e. the allocation of coordinates and filling of the solution arrays are now done in Kokkos loop on ```DefaultHostExecutionSpace```. Other minimal fixes include guards against undefined Compact Object Tracker parameters, that can lead to segfaults after interpolation onto the AthenaK grid, etc.

## Slow start lapse
Hengrui implemented a slow start lapse in another branch, that is based on https://journals.aps.org/prd/pdf/10.1103/PhysRevD.110.064045. He never pushed it, but I think this is a really helpful feature, so I took the liberty to put it on this branch. Since the z4c boosted puncture regression test now actively uses the SSL, the Theta norm is slightly higher, which is why I had to adjust the threshold by ~2%. 

## For reviewers
The horizon finder has been tested for all applications (SP, SSP, BBH, BHNS) and works well in all cases. The only thing that needs revision is the reordering of the Z4c tasklist. To get smooth excision to work, the Compact Object Tracker update needs to happen before the horizon finder computation, which in turn needs to happen before the C2P. Therefore, the Tracker and horizon tasks were moved into the main body of computation to ensure that dependency. Please check if that might collide with some other parts. From my side, I have not experienced any issues.

### DIFF
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -70,6 +70,15 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
         } else if (emethod.compare("lapse") == 0) {
           coord_data.excision_scheme = ExcisionScheme::lapse;
           coord_data.excise_lapse = pin->GetOrAddReal("coord","excise_lapse", 0.25);
+        } else if (emethod.compare("horizon") == 0) {
+          if (pin->DoesBlockExist("ahf")) {
+            coord_data.excision_scheme = ExcisionScheme::horizon;
+          } else {
+            std::cout << "### FATAL ERROR in " << __FILE__ << " at line "
+                    << __LINE__ << std::endl
+                    << "Horizon excision needs <ahf> block!" << std::endl;
+            std::exit(EXIT_FAILURE);
+          }
         } else {
           std::cout << "### FATAL ERROR in " << __FILE__ << " at line "
                     << __LINE__ << std::endl
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -73,6 +73,12 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
         } else if (emethod.compare("horizon") == 0) {
           if (pin->DoesBlockExist("fastflow")) {
             coord_data.excision_scheme = ExcisionScheme::horizon;
+            coord_data.smooth_excision = pin->GetOrAddBoolean("coord","smooth_excision",false);
+            coord_data.horizon_factor = pin->GetOrAddReal("coord","horizon_factor",1.0);
+            if (coord_data.smooth_excision) {
+              coord_data.texcise = pin->GetReal("coord","texcise");
+              coord_data.tdamp = pin->GetOrAddReal("coord","tdamp",1.0);
+            }
           } else {
             std::cout << "### FATAL ERROR in " << __FILE__ << " at line "
                     << __LINE__ << std::endl
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -71,12 +71,12 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
           coord_data.excision_scheme = ExcisionScheme::lapse;
           coord_data.excise_lapse = pin->GetOrAddReal("coord","excise_lapse", 0.25);
         } else if (emethod.compare("horizon") == 0) {
-          if (pin->DoesBlockExist("ahf")) {
+          if (pin->DoesBlockExist("fastflow")) {
             coord_data.excision_scheme = ExcisionScheme::horizon;
           } else {
             std::cout << "### FATAL ERROR in " << __FILE__ << " at line "
                     << __LINE__ << std::endl
-                    << "Horizon excision needs <ahf> block!" << std::endl;
+                    << "Horizon excision needs <fastflow> block!" << std::endl;
             std::exit(EXIT_FAILURE);
           }
         } else {
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -177,7 +177,7 @@ void Coordinates::CoordSrcTerms(const DvceArray5D<Real> &prim, const EOS_Data &e
     tt[2][3] = wtot * u2 * u3 + ptot * gupper[2][3];
     tt[3][3] = wtot * u3 * u3 + ptot * gupper[3][3];
 
-    // compute derivates of metric.
+    // compute derivatives of metric.
     Real dg_dx1[4][4], dg_dx2[4][4], dg_dx3[4][4];
     ComputeMetricDerivatives(x1v, x2v, x3v, flat, spin, dg_dx1, dg_dx2, dg_dx3);
 
@@ -321,7 +321,7 @@ void Coordinates::CoordSrcTerms(const DvceArray5D<Real> &prim,
     tt[2][3] = wtot * u2 * u3 + ptot * gupper[2][3] - b2 * b3;
     tt[3][3] = wtot * u3 * u3 + ptot * gupper[3][3] - b3 * b3;
 
-    // compute derivates of metric.
+    // compute derivatives of metric.
     Real dg_dx1[4][4], dg_dx2[4][4], dg_dx3[4][4];
     ComputeMetricDerivatives(x1v, x2v, x3v, flat, spin, dg_dx1, dg_dx2, dg_dx3);
 
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -56,8 +56,10 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
     if (coord_data.bh_excise) {
       // Set the density and pressure to which cells inside the excision radius will
       // be reset to.  Primitive velocities will be set to zero.
-      coord_data.dexcise = pin->GetOrAddReal("coord","dexcise",(FLT_MIN));
+      coord_data.dexcise = pin->GetReal("coord","dexcise");
+      coord_data.texcise = pin->GetReal("coord","texcise");
       coord_data.pexcise = pin->GetOrAddReal("coord","pexcise",(FLT_MIN));
+
       coord_data.flux_excise_r = (pin->DoesBlockExist("radiation")) ?
         1.0+sqrt(1.0-SQR(coord_data.bh_spin)) :
         pin->GetOrAddReal("coord","flux_excise_r",1.0);
@@ -79,7 +81,6 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
                                                               false);
             coord_data.horizon_factor = pin->GetOrAddReal("coord","horizon_factor",1.0);
             if (coord_data.smooth_excision) {
-              coord_data.texcise = pin->GetReal("coord","texcise");
               coord_data.tdamp = pin->GetOrAddReal("coord","tdamp",1.0);
             }
           } else {
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -57,8 +57,11 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
       // Set the density and pressure to which cells inside the excision radius will
       // be reset to.  Primitive velocities will be set to zero.
       coord_data.dexcise = pin->GetReal("coord","dexcise");
-      coord_data.texcise = pin->GetReal("coord","texcise");
-      coord_data.pexcise = pin->GetOrAddReal("coord","pexcise",(FLT_MIN));
+      if (is_dynamical_relativistic) {
+        coord_data.texcise = pin->GetReal("coord", "texcise");
+      } else {
+        coord_data.pexcise = pin->GetReal("coord", "pexcise");
+      }
 
       coord_data.flux_excise_r = (pin->DoesBlockExist("radiation")) ?
         1.0+sqrt(1.0-SQR(coord_data.bh_spin)) :
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -74,7 +74,8 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
         } else if (emethod.compare("horizon") == 0) {
           if (pin->DoesBlockExist("fastflow")) {
             coord_data.excision_scheme = ExcisionScheme::horizon;
-            coord_data.smooth_excision = pin->GetOrAddBoolean("coord","smooth_excision",false);
+            coord_data.smooth_excision = pin->GetOrAddBoolean("coord","smooth_excision",
+                                                              false);
             coord_data.horizon_factor = pin->GetOrAddReal("coord","horizon_factor",1.0);
             if (coord_data.smooth_excision) {
               coord_data.texcise = pin->GetReal("coord","texcise");
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -5,9 +5,10 @@
 //========================================================================================
 //! \file coordinates.cpp
 //! \brief
+#include <float.h> // FLT_MIN
+
 #include <iostream> // cout
 #include <string>
-#include <float.h> // FLT_MIN
 
 #include "athena.hpp"
 #include "mesh/mesh.hpp"
--- a/src/coordinates/coordinates.cpp
+++ b/src/coordinates/coordinates.cpp
@@ -7,6 +7,7 @@
 //! \brief
 #include <iostream> // cout
 #include <string>
+#include <float.h> // FLT_MIN
 
 #include "athena.hpp"
 #include "mesh/mesh.hpp"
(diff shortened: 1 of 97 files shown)


src/dyn_grmhd/dyn_grmhd.cpp
```diff
@@ -604,6 +614,15 @@ void DynGRMHDPS<EOSPolicy, ErrorPolicy>::AddCoordTermsEOS(const DvceArray5D<Real
...
     //Real E = tmunu.E(m,k,j,i);
 
+    // D = rho*W and tau = E-D are needed for the smooth damping terms
+    // inside excised regions, if existing.
+    Real D = prim(m, IDN, k, j, i) * W;
+    Real tau = E - D;
+
+    // Compute the excised value for the energy.
+    Real tau_ex = (dexcise*eos_.GetEnthalpy(dexcise/mb, texcise, &prim_pt[PYF]))
+                  + Bsq - pexcise - 0.5*bsq - dexcise;
```
jfields7: We should not use `pexcise` if `dexcise` and `texcise` are also defined because `pexcise` may no longer be constant or consistent. Since `dexcise` is computed assuming $W=1$, the computation for `tau_ex` can thus be simplified to the following:
```c++
    // Compute the excised value for the energy
    Real tau_ex = eos_.GetEnergy(dexcise/mb, texcise, &prim_pt[PYF]) + Bsq - 0.5*bsq - dexcise;
```

src/eos/primitive_solver_hyd.hpp
```diff
@@ -419,27 +420,32 @@ class PrimitiveSolverHydro {
...
+            prim_pt[PVY] = 0.0;
+            prim_pt[PVZ] = 0.0;
+            prim_pt[PPR] = pexcise_;
+            for (int n = 0; n < nscal; n++) {
+              // FIXME: Particle abundances should probably be set to a
+              // default inside an excised region.
+              prim_pt[PYF + n] = cons_pt[CYD]/cons_pt[CDN];
+            }
+            prim_pt[PTM] =
+              eos_.GetTemperatureFromP(prim_pt[PRH], prim_pt[PPR], &prim_pt[PYF]);
```
jfields7: Since `texcise` is now defined, it would be much better for even the non-smoothed excision to use it (see #634). This doesn't make a big difference for an ideal gas, but it's a huge difference for `EOSCompOSE` where defining `pexcise` is nontrivial, and `GetTemperatureFromP` is very expensive while `GetPressure` is relatively cheap.

src/pgen/elliptica.cpp
```diff
@@ -114,42 +153,37 @@ void SetupBinary(ParameterInput *pin, Mesh* pmy_mesh_) {
...
-        Real y = CellCenterX(j - js, nx2, x2min, x2max);
-        for (int i = 0; i < ncells1; i++) {
-          Real x = CellCenterX(i - is, nx1, x1min, x1max);
-
-          x_coords[idx] = x;
-          y_coords[idx] = y;
-          z_coords[idx] = z;
+  // Populate coordinates for Elliptica.
+  Kokkos::parallel_for("pgen_coordinates",
+  Kokkos::RangePolicy<Kokkos::DefaultHostExecutionSpace>(0, width),
```
jfields7, line 160: Does this work when compiling with both OpenMP and CUDA? I had issues in the past trying to do this very thing.

src/pgen/elliptica.cpp
jfields7, line 1: Since the file is now generic and can handle single neutron stars, BNS, and BHNS, it might be worth considering changing the filename to `pgen/elliptica.cpp` rather than `pgen/elliptica_bns.cpp`.

src/eos/primitive-solver/ps_types.hpp
```diff
@@ -24,6 +24,11 @@ enum ConsIndex {CDN=0, CSX=1, CSY=2, CSZ=3, CTA=4, CYD=5, NCONS=(NHYDRO)};
 // Note that NPRIM does not include IBY and IBZ because NHYDRO doesn't.
 enum PrimIndex {PRH=0,PVX=1, PVY=2, PVZ=3, PPR=4, PTM=5, PYF=6, NPRIM=((NHYDRO)+1)};
 enum SpatialMetricIndex{S11=0, S12=1, S13=2, S22=3, S23=4, S33=5, NSPMETRIC=6};
+enum ExtrinsicCurvatureIndex{K11=0, K12=1, K13=2, K22=3, K23=4, K33=5, NEXCURV=6};
+enum SpatialMetricDrvsIndex{D1S11=0, D1S12=1, D1S13=2, D1S22=3, D1S23=4, D1S33=5,
+                            D2S11=6, D2S12=7, D2S13=8, D2S22=9, D2S23=10, D2S33=11,
+                            D3S11=12, D3S12=13, D3S13=14, D3S22=15, D3S23=16, D3S33=17,
+                            NDRVSSPMETRIC=18};
```
jfields7: There's nothing in PrimitiveSolver that depends on these quantities, so I'm not sure this file is the best place for these additions.

tst/test_suite/z4c/test_z4c_boosted_gpu.py
```diff
@@ -19,7 +19,7 @@
     ("Mx-norm"): (1.1e-03),
     ("My-norm"): (4.5e-04),
     ("Mz-norm"): (4.5e-04),
-    ("Theta-norm"): (3.1e-05),
+    ("Theta-norm"): (3.15e-05),
```
jfields7: Is SSL on by default? If this has to be adjusted because of SSL, maybe we should make sure it's off for the regression test. I don't know if other people have opinions on this, though.

jfields7 (changes requested):
Thanks for all your work on this! These will be very useful to have in the code. I admit I'm not 100% familiar with how the horizon finder works, so most of my comments concern the smooth excision. I do have a couple additional comments not covered in my suggestions below.
1. It would be very useful to have some regression tests for the horizon finder. Maybe something can be incorporated into the existing boosted puncture test?
2. I feel like I've asked this question before, but what is the logic behind having a new implementation of Lagrange interpolation? We already have a Lagrange interpolator used for the compact object trackers, and I think we should resist duplicating code that does the same thing if at all possible.

tst/test_suite/z4c/test_z4c_boosted_gpu.py
spacetimecurv: Yes, SSL is on by default. I assume Hengrui forgot to push the implementation of the SSL to the main branch. Still it is used in the regression test but does not do anything. I turned it off (which I think is better) and will push the change with the old Theta-norm value once the rest of the comments are treated by me.

src/eos/primitive-solver/ps_types.hpp
spacetimecurv: You are right, I put this inside the horizon finder class and will also push this.

src/pgen/elliptica.cpp
spacetimecurv: Good idea!

src/pgen/elliptica.cpp
spacetimecurv: I only checked CUDA only, and there it works. But I can also check both OMP and CUDA. This implementation is taken from Alan's Kadath pgen, where he basically does the same. I assume he has tested this with both OMP and CUDA @alanlam1002?

spacetimecurv:
To your above questions: 
1. I tried including the horizon finder in the boosted puncture regression test, which luckily reveiled an interesting bug, that I resolved (the metric derivatives array needs to be reallocated if the number of MBs changes per rank, due to the AMR grid moving across the domain). To get the horizon finder to work at the initial time step, I needed to change the initial refinement region in x from [-1,0] to [-1,1], because otherwise the other half of the puncture would be underresolved and the horizon finder fails at the first step. Now if these changes are fine, then I can add this to the boosted puncture test, otherwise I can add a separate regression test for the horizon finder. Thoughts?
2. The reason is mainly performance. Our main horizon finder workflow is `Compute metric derivatives`->`Interpolate metric onto the trial surface`->`Compute the surface integrals`. In this workflow the biggest bottleneck is the interpolation. This scales badly for large numbers of angular points on the trial surface. Since the horizon finder is used online during the simulation, this can result in drastic performance drop-offs. Putting this inside a GPU kernel requires a device-safe interpolator, which the Lagrange Interpolator class is not. This is why I mirrored this to be device-safe inside an inline function. This is not the best work-around, but I do not see a better option. Otherwise we would have to call thousands of Lagrange Interpolators at each time step in serial or host parallelized loops.

src/dyn_grmhd/dyn_grmhd.cpp
spacetimecurv: This is now implemented.

src/eos/primitive_solver_hyd.hpp
spacetimecurv: Done, please check that this is correct.

src/pgen/elliptica.cpp
spacetimecurv: I checked a really low-resolution test with CUDA and OpenMP and had no issues. But I also do not know what troubles you ran into in the past.

jfields7: The code simply refused to compile, so as long as it compiles, it should be fine.

src/coordinates/coordinates.cpp
```diff
@@ -55,7 +57,9 @@ Coordinates::Coordinates(ParameterInput *pin, MeshBlockPack *ppack) :
       // Set the density and pressure to which cells inside the excision radius will
       // be reset to.  Primitive velocities will be set to zero.
       coord_data.dexcise = pin->GetReal("coord","dexcise");
-      coord_data.pexcise = pin->GetReal("coord","pexcise");
+      coord_data.texcise = pin->GetReal("coord","texcise");
+      coord_data.pexcise = pin->GetOrAddReal("coord","pexcise",(FLT_MIN));
```
jfields7: The static spacetime HARM-like GRMHD solver still relies on pressure, so these need to be gated behind the correct physics modules. Something like the following:
```c++
if (is_dynamical_relativistic) {
  coord_data.texcise = pin->GetReal("coord", "texcise");
} else {
  coord_data.pexcise = pin->GetReal("coord", "pexcise");
}
```

tst/inputs/rad_beam.athinput
```diff
@@ -63,6 +63,7 @@ a           = -0.9    # black hole spin a (0 <= a/M < 1)
 excise      = true   # excise black hole
 dexcise     = -1.0   # excision density
 pexcise     = -1.0   # excision pressure
+texcise     = -1.0   # excision temperature
```
jfields7: With the requested changes elsewhere, this isn't needed. The radiation solver doesn't currently work with `DynGRMHD`, which is the one that needs `texcise`.

jfields7 (changes requested):

jfields7 (approved):

HengruiZhu99:
Overall looks good to me! A few things clean ups I'd suggest before merging.

1. In `PrimitiveSolverHydro`, the non-smoothed excision branch now always uses `texcise` to compute the pressure. Since `coordinates.cpp` only initializes `texcise` for dynamical-relativistic runs and keeps `pexcise` for the static/HARM-like GRMHD path, this looks like it may break existing static excision cases. Also, `GetPressure(..., &prim_pt[PYF])` is called before `prim_pt[PYF+n]` is filled, which is unsafe for composition-dependent/tabulated EOS. I think we should preserve the old `pexcise` path for non-dynamical/static excision, and fill `PYF` before any EOS call in the temperature-based path.

2. In `DynGRMHDPS::AddCoordTermsEOS`, `tau_ex` is computed unconditionally, even when `smooth_excision` is false. Since this uses `texcise`/excision target state and makes an EOS call, I think it should be moved inside the `if (smoothing)` block.

3. `FastFlow::ah_found` and `time_first_found` should probably be explicitly initialized in the constructor. `Find()` can return early before `FastFlowLoop()` sets `ah_found = false`, while horizon excision later reads `ah_found` and `rr_min` to build the mask. That could make the mask depend on uninitialized state before `start_time` or while waiting for the punctures to be close.

A couple smaller points: I would either clarify/test the `store_metric_drvts=true` path, since `MetricDerivatives()` returns early while `MetricInterp()` still reads `dg_`, or disable/remove that option for now. Also, `athena_read.horizon()` should probably use `np.atleast_2d` so it does not fail on a single-row horizon summary file.

jfields7:
I don't think the `texcise` vs. `pexcise` question is a big issue. The `is_dynamical_relativistic` flag is toggled when `DynGRMHD` is available, not only if Z4c is enabled, and `PrimitiveSolverHydro` is only used by `DynGRMHD`. The HARM-like solver has a similar but much less general primitive recovery routine that only checks for `pexcise`. `DynGRMHD`, on the other hand, should only use `texcise`; finding an excision pressure that works for anything other than an ideal gas is a pain in the neck. As maybe a last comment on this, the regression suite includes a GR Bondi accretion test, and that passes, so this should be fine.

Calling the pressure before initializing the scalars is unlikely to cause issues in production simulations because `EOSCompOSE` clamps the inputs to physical values, but this should still be fixed.

Another small thing I noticed: smooth excision is only enabled for horizon-based masking, but there's no reason it shouldn't also work for lapse-based masking. I'm not sure if it works with fixed masking because Kerr-Schild doesn't excise the curvature singularity the way that the puncture gauge does, but it might work there, too. Since this needs additional testing, it could be delayed to a later pull request, but it's worth thinking about.

spacetimecurv:
@HengruiZhu99 I addressed all of your points from above, except the one that Jacob justified. On your third point, I just added ```time_first_found```, ```ah_found``` etc. to the constructor with a commit today, because with data from checkpoints we should use the last found surface as well. Otherwise, we use the "bad" guess in isotropic coordinates. Also, the ```use_stored_metric_drvts``` flag was a left-over from GR-Athena++ and only makes sense in static spacetimes. I commented it out for now, since this should only rarely be used.

@jfields7 I agree, we could also use smooth excision for lapse-based and maybe also for fixed excision. Once I find the time, I can investigate this separately and make some tests.

dradice (approved):

spacetimecurv:
The last commit enabled smooth excision to be used for lapse-based and fixed excision in dynamical GRMHD. Tests involving a single puncture embedded in a constant fluid background have worked with these methods as well. Furthermore, the Lorene/Elliptica pgens now have an additional guard on the electron fraction, i.e. if someone uses a tabulated EOS, then the ```TOV::UsesYe``` function will return true but if ```nscalars``` is not set, then this can lead to weird overflow behaviors in memory and corruption of the initial data. In this fix we make sure to only get the electron fraction from the density if ```nscalars > 0```. This is now the last thing that was left to do. If no one has a complaint, then this can be merged.

PR #747, opened 2026-06-11
Add -v input-file validation and unused-parameter warnings

dradice:
Add a "-v" command-line option that constructs the Mesh, problem generator and all physics classes -- exercising every parameter read, the initial-data reader and the EOS table -- and then quits before the evolution loop, so an input file can be validated against the actual code paths before submitting a job. To keep this cheap, -v collapses the root grid to a single MeshBlock and disables mesh refinement.

Also track whether each input parameter is read (InputLine::accessed, set in GetPtrToLine; code-inserted GetOrAdd*/Set* entries are flagged via a new from_code argument to AddParameter) and warn on stderr about any unused parameters after all classes are constructed, in every run. This catches typos in optional parameters, which are otherwise silently ignored. Warnings print only from rank 0; <comment> blocks (and, under -v, mesh-refinement blocks) are skipped.


jmstone:
This looks useful, but there are conflicts now with the main branch.  If those can be fixed, I am happy to merge.  Probably should add something in the docs about this too.

dradice:
> This looks useful, but there are conflicts now with the main branch. If those can be fixed, I am happy to merge. Probably should add something in the docs about this too.

It should merge cleanly now (there was just a whitespace issue before).

PR #748, opened 2026-06-12
Fix refinement at boundary for Z4c/MHD/Hydro/Radiation

### LINKED ISSUE #738: Z4c AMR at domain boundaries
There is a longstanding issue in the code where the Z4c variables are filled with garbage if the AMR refines or derefines a mesh block at the physical mesh boundary. Most production runs use grids large enough that we don't need to worry about refinement boundaries touching the edge of the domain, but it makes small-scale and low-resolution tests difficult, and it also means that bitant/reflection symmetry cannot be used in runs with Z4c and AMR.

spacetimecurv:
A long-standing issue in the Z4c sector has been refinement at the boundary, see #738. After a longer investigation it turns out, that this issue is not confined to Z4c, but appears in other physics modules as well. The fix is simple and touches as little of the codebase as possible, as will be explained below. 

## Z4c
To trigger refinement at the boundary we take a TOV star with an initially centered refinement level with tracker AMR and center the tracker in one corner of the grid, such that after the first refinement we get a configuration of the form:
tov mhd_w_bcc 00000

This results in a fine/coarse junction at the boundary. Mapping the points where $\chi \leq 0$ on the coarse/fine (blue/red) grid with C2P errors shown in green, we find:
amr_view_00_e20_a30

These points sit at the interface, where a fine/coarse block sit at the boundary. What happens is the following:
1. We apply the physical boundary conditions to the newly refined blocks and in the Z4c tasklist to the coarse/fine grid.
2. We then prolongate from coarse to fine to fill the ghost zones.

However, at these junctions where a coarse and a fine block meet at the boundary, we require that the ghost zones of the fine grid are filled before the boundary conditions are applied. Otherwise we apply the boundary conditions to unfilled points and this results in C2P errors later on. The simple fix is to split the BC application to the coarse and fine grid and re-order the operations:
1. Apply the boundary conditions to the coarse grid.
2. Prolongate from coarse to fine.
3. Apply the boundary conditions to the fine grid.

This is consistent with what is done in GR-Athena++ and solves all of these problems. @jfields7 had a different setup, where he applies a custom refinement condition to a TOV star that emulates a BNS at a certain separation. In that case the garbage from the boundary can propagate inwards and cause C2P errors on random positions on the grid. This is solved by this fix.

## MHD/Hydro
Consequently, we also find this problem in the MHD/Hydro sector, where it might not be too obvious. A simple test was to take a blast wave setup with reflecting boundary conditions:
https://github.com/user-attachments/assets/3550464c-626e-42fc-86de-2be60e88ea49

The system is not stable but produces this fine/coarse junction on the top/bottom boundary. At these regions we again get C2P errors:
view_04

By applying the same trick as for the Z4c sector here, these C2P errors vanish. To check that we still conserve the divergence-free constraint, the next video shows the divergence of the magnetic field for the same test with an outflowing boundary:
https://github.com/user-attachments/assets/f842ed37-824e-4b29-8151-b2984be13513

## Radiation
The same fix as above, also has to be applied to the Radiation sector using the fixes for the MHD/Hydro sector from above when running Radiation with MHD/Hydro.

## For discussion
This fix completely resolves all problems at the boundary. It is definitely needed in the Z4c sector and it is also the correct thing to do for the other physics modules, although the impact here is less clear. Further, we barely touch the code base and only re-order operations, which is consistent with GR-Athena++. This should not have a large impact on performance albeit we now also have to apply the boundary conditions to the coarse grid in the MHD/Hydro/Radiation sector to correctly fill fine blocks at the boundary. Other ideas are welcome!


HengruiZhu99:
Thanks for tracking this down — the overall fix makes sense to me. Reordering the AMR boundary handling as

1. apply physical BCs on the coarse array,
2. prolongate coarse → fine,
3. apply physical BCs on the fine array,

seems like the right approach for the fine/coarse-at-physical-boundary issue, and it directly addresses the failure mode in #738. The corresponding task-list changes for Z4c, Hydro/MHD, Radiation, and DynGRMHD also look consistent to me.

I do have two concerns before merging:

1. **User-defined BCs with AMR at physical boundaries.**  
   As far as I can tell, the new coarse-BC path only applies the built-in BC helpers. If a boundary is `BoundaryFlag::user`, those helpers just fall through, while `user_bcs_func` is still only called later in `ApplyPhysicalBCs()`, i.e. after prolongation. So for AMR blocks touching a physical boundary with user BCs, the prolongation stencil may still read unfilled coarse ghost zones. Could we either apply user BCs on the coarse data before prolongation as well, introduce a coarse-aware user-BC path, or explicitly disallow/warn for this combination?

2. **The `ChangeRunDir(run_dir)` move seems unrelated.**  
   Moving this before `AddCoordinatesAndPhysics()` / `ProblemGenerator` construction changes the working directory seen by physics constructors and pgens. That may be useful, but it is orthogonal to the BC/prolongation fix and could change relative-path behavior. I would prefer to split that into a separate PR unless it is required here and documented.

It would also be good to add a small regression test for this specific AMR-at-domain-boundary configuration, if practical, so we do not accidentally reintroduce the ordering issue later.

Aside from those points, the core changes look correct to me.

spacetimecurv:
1. You are right, this is not covered. There is a simple option to add a ```coarse=True/False``` flag to the ```user_bcs_func``` function pointer. However, the bigger problem is that many of these user-defined boundary conditions, such as the one for the torus use C2P inversions etc. that operate on the fine array. In other words there needs to be a lot of change for each user-defined boundary function to distinguish between coarse/fine. I do not think this is do-able without major changes. For now I would just place a warning in the code that tells the user to try and avoid refinement at the boundary. If you have other suggestions, then I would appreciate them.
2. I played around with this and forgot to revert the change. While this is nice to have for local tests, it is not needed when using batchtools or other software. This should not break anything, it only leads to files that are created in the constructor of different physics modules, e.g. the CompactObjectTracker, to be created inside the run directory and not on the level of the program call. Anyways, I reverted this change.

I will also add a small regression test and push this as soon as it is done.

HengruiZhu99:
Looks good to me! @jmstone ready to be merged

jfields7 (commented):
Overall I think this looks fine. I agree that a regression test would be helpful, and once that and the privately discussed mass conservation plot are ready, I'm happy to approve this.

spacetimecurv:
First of all, in the following I am sharing the requested mass conservation for the above Blast wave test with outflowing boundary conditions:
mass_cons

The good thing is, that the fix does not brake mass conservation, because the violation at the end is when material is leaving the domain. On the other hand, it looks like that due to the refinement at the boundary without the fix we actually get an increased violation in mass conservation due to the accumulation of C2P errors.

I have added two regression tests, one of a linear wave in GRMHD, where we check the C2P failures which should stay 0 and one of a linear wave in Z4c, where we look at the errors. It turns out that for 2 ghost cells the errors are not affected too much whether the fix is on or not, but for 4 ghosts the difference is quite large (black solid and red dashed curves):
lwavez4_norm

For the linear wave in MHD, the difference is minor:
lwave_norm

jfields7 (approved):

dradice (approved):

jmstone (approved):
Looks good, and may actually fix problems occuring in MHD when AMR reached physical boundaries that have emerged in other applications.

PR #754, opened 2026-06-27
Fix derived temperature output

Haocheng-fdu:
This fixes the derived `temperature` output.

The previous implementation computed:

pressure(i+1) / density(i-1)

which is inconsistent with the comment temperature = pressure / density and can introduce ghost-zone or meshblock-boundary artifacts.

This PR changes the derived temperature calculation to use same-cell pressure and density:

pressure(i) / density(i)

It also selects the hydro or MHD primitive array based on which pack is available, rather than checking an unrelated derived-variable name.


jmstone:
Thanks for flagging this, it was actually fixed (along with some other bugs in the derived variables function) in another PR which was just merged.

PR #757, opened 2026-06-29
Split recon rsolver clean

davidvelasco07:
# Split reconstruction + Riemann-solve scheme:

This note summarizes the performance trade-offs of the **split** flux scheme
(separate reconstruction and Riemann-solver kernels) versus the **baseline**
monolithic flux kernel, using three benchmark sets:

1. **Hydro linwave** throughput — a digitized GH200 sweep (LLF solver), plus a
   directly measured single-A100 sweep (HLLC solver).
2. **MHD linwave** throughput — a directly measured single-A100 sweep (HLLD).
3. **dyngr-TOV** single-GPU throughput on GH200 (wenoz+hlle).

---

## What the split scheme does

The baseline computes fluxes in a single, monolithic kernel: for every face it
reconstructs the left/right states *and* immediately calls the Riemann solver,
keeping the reconstructed states in registers.

The split scheme breaks this into two stages:

- **Reconstruction kernel** — computes the left/right primitive states at every
  face and writes them to global per-face buffers (`wl3d`, `wr3d`, and
  for MHD `bl3d`, `br3d`, …). Method selection (PLM/PPM/WENOZ/…) is
  resolved at compile time via templated dispatch (`ReconCellT`, `ReconDispatch`).
- **Riemann-solve kernel** — reads those buffers and computes the numerical
  flux with the chosen solver (`LLF`, `HLLE`, `HLLC`, `Roe`, …).

The key structural difference is *where the reconstructed states live*: in
registers (baseline) versus in global memory (split).

---

### Hydro linwave, GH200 (LLF)

| MeshBlocks | edge 64 | edge 128 | edge 256 | edge 512 |
|-----------:|:-------:|:--------:|:--------:|:--------:|
| 1          | 2.00x   | 2.06x    | 1.55x    | 0.92x    |
| 64         | 1.26x   | 1.15x    | 0.96x    | 0.93x    |
| 4096       | 1.36x   | 1.43x    | 1.37x    | 1.15x    |

(speedup = split / baseline; >1 means split is faster)

### Hydro linwave, single A100 (HLLC)

Measured on one NVIDIA A100-PCIE-40GB (HLLC+PLM, ideal EOS, 200 fixed
cycles). edge 512 is omitted (OOMs on the 40 GB card for both builds).

| MeshBlocks | edge 64 | edge 128 | edge 256 |
|-----------:|:-------:|:--------:|:--------:|
| 1          | 1.97x   | 1.23x    | 0.95x    |
| 64         | 1.59x   | 1.30x    | 1.04x    |
| 4096       | 1.32x   | 1.53x    | 1.54x    |

(speedup = split / baseline; >1 means split is faster)

### MHD linwave, single A100 (HLLD)

Measured on one NVIDIA A100-PCIE-40GB (HLLD+PLM, ideal EOS, 200 fixed
cycles). edge 512 is omitted (OOMs on the 40 GB card for both builds).

| MeshBlocks | edge 64 | edge 128 | edge 256          |
|-----------:|:-------:|:--------:|:-----------------:|
| 1          | 1.77x   | 1.45x    | n/a¹              |
| 64         | 2.41x   | 2.13x    | 1.55x             |
| 4096       | 2.27x   | 2.85x    | 2.76x             |

(speedup = split / baseline; >1 means split is faster)

¹ At edge 256 / 1 MeshBlock the **baseline** aborts with a Kokkos `TeamPolicy`
"could not find a valid execution configuration" launch error (per-team scratch
too large for one 256³ block); the split build runs it fine (0.213 GZCPS), so no
ratio is defined.

### dyngr-TOV single-GPU, GH200 (wenoz+hlle)

dyngr-TOV: baseline vs split

| MeshBlocks | edge 64 | edge 128 | edge 256 |
|-----------:|:-------:|:--------:|:--------:|
| 1          | 4.77x   | 3.73x    | 1.23x    |
| 64         | 1.49x   | 1.60x    | 1.56x    |
| 4096       | 1.50x   | 1.67x    | 1.62x    |

(speedup of `split (recon+rsolve)` over baseline)

---

## Pros

- **Large gains for heavy, register-hungry physics.** The dyngr-TOV
  (GR-MHD) flux kernel has very high register pressure, which throttles GPU
  occupancy in the monolithic form. Splitting it into two smaller kernels
  relieves that pressure and yields **1.5x–4.8x** throughput, with the biggest
  wins at small domains where occupancy matters most.
- **Consistent wins in the small-per-kernel-work regime.** Whenever the
  baseline kernel is occupancy-limited — single small MeshBlock (up to ~2x for
  hydro), or many small blocks (4096 MeshBlocks: ~1.15x–1.67x) — the smaller
  split kernels schedule better and run faster.
- **Cleaner separation of concerns.** Reconstruction and the Riemann solve are
  independent kernels, so each can be tuned, profiled, and launched
  independently. Compile-time templated dispatch keeps the per-method code
  monomorphic and avoids runtime branching in the hot loop.
- **Easier to extend.** New reconstruction methods or solvers plug into the
  dispatch layer without touching a single giant kernel.

## Cons

- **Extra global-memory traffic.** The reconstructed left/right states are
  written to global buffers and then re-read by the solver kernel, instead of
  staying in registers. This is pure added bandwidth cost.
- **Slight regressions when the baseline is already efficient.** For workloads
  that already achieve good occupancy — e.g. hydro with 64 MeshBlocks at large
  domains (edge 256: 0.96x, edge 512: 0.93x) and the single large block at edge
  512 (0.92x) — the extra buffer traffic is not amortized by an occupancy gain,
  so split is marginally slower.
- **Higher memory footprint.** The global per-face state buffers (`wl_split`,
  `wr_split`, and the MHD/field equivalents) must be allocated, increasing
  device memory use.
- **More kernel launches.** Two launches per flux evaluation instead of one;
  the per-launch overhead is small but nonzero.

---

In short: the split scheme trades a small amount of extra memory bandwidth for
a large reduction in register pressure. That trade is strongly favorable for
heavy kernels and small-work regimes, roughly neutral-to-slightly-negative for
already-saturated, lightweight kernels.


jmstone (approved):
Thanks, this is certainly a major change.  It reverts to the very first version of the hydro flux functions implemented in 2020 which also used split kernels.  When the code was simple it turned out fused kernels was better.  But now with all the different reconstruction and Rsolvers you have made a convincing case this split version is better, plus it is more flexible.  So I am happy to merge this.  I note M. Guo and others have tested this on their applications and find it works.  But all users of the code should be aware this is a major change and report if they have any problems.

PR #758, opened 2026-06-30
Aimbipolar

callumf8:
Title: Add ambipolar diffusion (non-ideal MHD) to the Resistivity module

## Summary

Adds ambipolar diffusion (AD) to AthenaK's MHD solver, following the Athena++
`field_diffusion` algorithm. AD is folded into the existing `Resistivity` class rather than
added as a new module, so the core-code footprint is small and ideal-MHD behaviour is
unchanged. A non-zero `<mhd>/eta_ad` enables it.

## What's added

- **Ambipolar EMF** at cell edges: `E_AD = eta_ad*[B^2 J - (J.B) B]` (the expanded form of
  `eta_A*[J - (J.B)B/B^2]` with `eta_A = eta_ad*B^2`, so no division by `B^2` / no floor).
- **Ambipolar Poynting flux** on the energy flux: `S_AD = eta_ad*B^2*(J x B)`, using
  Athena++'s "average-then-multiply" pattern.
- **Inline current density**: `J = curl(B)` is recomputed at the edges from face-centred `B`
  via `EdgeJ{1,2,3}` helpers (no stored `jedge` array); cross-components are averaged to the
  target edge (4-pt in 3D, 2-pt across a degenerate direction in 2D), matching Athena++
  `AmbipolarEMF`.
- **Per-cell parabolic timestep**: `dt = fac*min_cells,dirs[ dx^2/(eta_ohm + eta_ad*B^2_cell) ]`,
  pairing each cell's own `dx` with its own diffusivity -- equivalent to Athena++
  `FieldDiffusion::NewDiffusionDt` and AthenaK's `Conduction::NewTimeStep`.

## Design notes

- AD methods live in a dedicated `src/diffusion/ambipolar.cpp` but remain members of the
  `Resistivity` class (declared in `resistivity.hpp`). The Ohmic path and `NewTimeStep` stay
  in `resistivity.cpp`.
- **Off-by-default / no impact when disabled**: when `eta_ad == 0`, AD kernels are never
  dispatched and the timestep uses the original constant-coefficient path, so Ohmic-only and
  ideal-MHD runs are bit-identical to before. No changes to the task list, CT update, or
  global-dt machinery.

## Validation

- **Cross-code agreement**: ambipolar linear-wave damping (Bai & Stone 2011, Sec 2.3.2)
  reproduces Athena++ to 4 significant figures across all 9 cases (1D/2D/3D x
  fast/Alfven/slow).
- **Convergence to analytic**: damping rates match the Balsara (1996) analytic values to
  <2.5% at the "double" resolution.

## Testing

- New `tst/test_suite/diffusion/test_diffusion_ambipolar_linwave_{cpu,gpu}.py` (Bai & Stone
  wave damping). CPU: 1D & 2D at N=64 (+/-20%), 6 cases ~10 s. GPU: 2D & 3D at N=128 (+/-5%,
  measured 0.5-2.3%), 6 cases ~150 s.
- Verified builds and runs on both CPU (Serial) and GPU (CUDA, A100).


jmstone (approved):

jmstone:
Looks great, thanks Callum!

PR #760, opened 2026-07-01
Add internal dipole magnetic field to Elliptica mixed binaries

spacetimecurv:
This PR adds a pressure-poloidal magnetic field configuration to the Elliptica pgen for black hole-neutron star (BHNS) systems based on https://arxiv.org/pdf/2211.07158 (Eq. (1)). With this modification, we now have the possibility to evolve BHNS initial data with either an external or internal dipole field. For this new model we compute the maximum gas pressure, and based on this the vector potential. Since the computation happens with unit amplitude, we need to rescale to the desired field strength at the end. The resulting field looks like:
mag_field

Simulations with this field are ongoing and give reasonable results. Extension to binary neutron star (BNS) systems is possible, but the majority of simulations are conducted with the Lorene pgen. The logic can thus be transferred there if desired.

Note: the pgen guards against usage of this model with BNS or isolated neutron stars and throws a warning, that the finest refinement level should confine the entire neutron star. The reason is that otherwise we would need to interpolate the pressure to sub-edges at fine/coarse interfaces, which can lead to violations of the divergence-free constraint. Since this is almost always the case in production runs, this is only a minor restriction.


jfields7 (approved):
This looks good to me. Once the merge conflict is resolved, this is ready to be merged.

spacetimecurv:
> This looks good to me. Once the merge conflict is resolved, this is ready to be merged.

Done!

PR #768, opened 2026-07-17
updating pgen #includes to match new directory structure

ajdittmann:
I tried to compile the shu_osher problem earlier today and noticed that many of the header paths had not been updated to match the reorganization introduced in #761. 
I've tried to update all of the pgens accordingly.


jmstone (approved):
Very helpful, thanks.

PR #770, opened 2026-07-21
Import Drift-Control (DC) from THC/GR-Athena++

spacetimecurv:
This PR adds utility for the Drift-Control (DC) from THC/GR-Athena++. By enabling the DC, based on the method specified, the RHS of the Z4c equations for the shift $$\beta^i$$ get modified and result in a damped motion of the compact object. This is particularly useful for instance in black hole-neutron star simulations, where the kick of the black hole after merger will remove it from the grid origin. For long-term simulations this is desirable, since this also shifts the analysis of the neutrino luminosities or ejecta, since the extraction spheres stay static, while the entire system moves.

Supported are 3 methods: Oscillator (THC), PID and Relaxation (GR-Athena++)

This is the result of a displaced single puncture:
comp_OnePuncture
The target was the grid origin, and all methods show reasonable results, where the differences/deviations are result of parameter tuning. With parameter tuning and enabling/disabling the algorithm we can ensure better convergence.

For a boosted puncture with a fixed velocity we get:
comp_BoostedPuncture
PID performs the best, while the other two methods reduce the velocity at earlier times (see the plateau), but are not sufficient to completely stop the puncture. A more aggressive parameter setting would have likely achieved this goal.


dradice (approved):

HengruiZhu99 (commented):
Over all it looks good! Two issues I'd raise:

I think we should synchronize the drift-control position with the compact-object tracker on restart. Right now dc_pos is initialized to dc_fixed, while the tracker position is restored later, so the controller seems to see the wrong position for the first step after restart. 

Also, the control parameter is updated per RK timestep, not per stage, so the convergence would be reduced to first order (formally). Though changing it may also require changing the puncture tracker to be coupling to RK stages, and looking at the result, this might not dominate the error budget any way. So just a minor point, probably fine as it is given your tests.

spacetimecurv:
> Over all it looks good! Two issues I'd raise:
> 
> I think we should synchronize the drift-control position with the compact-object tracker on restart. Right now dc_pos is initialized to dc_fixed, while the tracker position is restored later, so the controller seems to see the wrong position for the first step after restart.
> 
> Also, the control parameter is updated per RK timestep, not per stage, so the convergence would be reduced to first order (formally). Though changing it may also require changing the puncture tracker to be coupling to RK stages, and looking at the result, this might not dominate the error budget any way. So just a minor point, probably fine as it is given your tests.

Thanks for the review. Regarding your issues:

1. I actually don't think that is a problem. Assume we want to pull a remnant black hole to the origin, and before we write the restart it moved from (-2.5, -2.5, 0.0) to (-0.5, -0.5, 0.0), then the tracker position will write the latter into the checkpoint. When reading from the checkpoint, as you said, driftcontrol will write the position to be (0.0, 0.0, 0.0). You then evolve the tracker by one stage and then evolve the driftcontrol with the updated tracker position and compute the error budget from that. So you are right to say that there is like a one-cycle lag between the two, but I am not sure it makes such a big difference, because you just "pay" a little bit more force to pull the black hole to the target at the first step.
2. You are right, this is not ideal but in most cases it should not matter. This is not ideal, but in my tests with kicked black holes from BHNS mergers all of those methods worked well, so I assume this should be okay.

HengruiZhu99:
Thanks for the clarification! @jmstone ready to be merged.

PR #772, opened 2026-07-22
Extend Spherical Surface Output class to multiple radii

spacetimecurv:
In the current implementation of AthenaK, if one desires to have `sph` output for a given variable, e.g. the density `dens`, at multiple radii then the natural approach is to add one `sph` output block for each radius. For things like a tracer analysis of outflowing material, where we require resolutions of $$N_r\times N_{\phi}\times N_{\theta}=128\times 256\times 128$$, so 128 radii, adding 128 output blocks for just one variable is a complete overkill. In GR-Athena++, this was solved in sort of a hacky way, but in AthenaK it is a natural extension of the `SphericalSurface` class and subsequently of `SphericalSurfaceOutput`. 

This PR adds this extension to multiple radii. The user can now specify the original single-radii output as before by specifying:
```
<outputx>
file_type = sph
variable = mhd_w_bcc
radius = 100.0
ntheta = 128
dt = 1.0
xc = 0.0
yc = 0.0
zc = 0.0
```
In addition to that, a list of radii can also be parsed by setting:
```
<outputx>
file_type = sph
variable = mhd_w_bcc
radius = 10.0, 20.0, 50.0
ntheta = 128
dt = 1.0
xc = 0.0
yc = 0.0
zc = 0.0
```
Lastly, a list of radii within some bounds can also be generated via:
```
<outputx>
file_type = sph
variable = mhd_w_bcc
r_min = 10.0
r_max = 100.0
nradii = 10
r_spacing = log # linear
ntheta = 128
dt = 1.0
xc = 0.0
yc = 0.0
zc = 0.0
```

The created VTK output therefore holds $$(r,\theta,\phi)$$ columns.

### DIFF
--- a/src/outputs/spherical_surface.cpp
+++ b/src/outputs/spherical_surface.cpp
@@ -10,10 +10,14 @@
 
 #include <sys/stat.h>  // mkdir
 
+#include <algorithm>
+#include <cmath>
 #include <cstdio>  // snprintf
 #include <fstream>
+#include <iostream>
 #include <sstream>
 #include <string>
+#include <vector>
 
 #include "athena.hpp"
 #include "globals.hpp"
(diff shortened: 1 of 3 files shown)


jfields7:
This looks good to me. Can you also make a pull request in the [docs repo](https://github.com/IAS-Astrophysics/athenak-docs) to update the documentation for this?

@jmstone, will any of these changes cause problems with your planned refactoring of the output system?

spacetimecurv:
@jfields7 @jmstone The updated documentation tailored to this PR is created under https://github.com/IAS-Astrophysics/athenak-docs/pull/8.

jfields7 (approved):

PR #775, opened 2026-07-30
Feature/rank packed bvals

davidvelasco07:
# Rank-packed boundary communications

This change replaces AthenaK's per-(MeshBlock, neighbour) MPI boundary exchange with a
**rank-packed** exchange: one aggregated message per peer rank per exchange, packed and
unpacked directly by the existing `SendBuff`/`RecvBuff` device kernels.

It is a pure communication-layer change. Results are bit-identical to `main` on uniform
and static-refinement meshes at every rank count tested.

---

## 1. Motivation

`MeshBoundaryValues::PackAndSendCC/FC` currently posts one `MPI_Isend` per
(MeshBlock, neighbour-buffer) pair, and `InitRecv` posts the matching `MPI_Irecv`. A 3D
MeshBlock has up to 56 neighbour buffers (6 faces + 12 edges + 8 corners, times the
sub-block multiplicity under SMR/AMR), so a rank owning `M` MeshBlocks posts up to
`56·M` messages per variable exchange — several times per RK stage.

Almost all of those messages are small. For a 16³ block, a face buffer at `nvar = 5`
carries 5·2·16·16 = 2560 doubles ≈ 20 KB; an edge buffer carries ≈ 2.5 KB and a corner
buffer ≈ 320 B. At these sizes the exchange is entirely **latency-bound**: wall time
scales with the message *count*, not the byte count.

Two trends make this worse rather than better:

- **Small blocks are desirable.** AMR, load balance granularity, and GPU occupancy all
  push toward many small MeshBlocks per rank. That is exactly the regime where the
  message count explodes.
- **GPU density per NIC is rising.** Current dense-GPU nodes put several GPUs behind a
  single network interface — a Vista B200 node has four. Four ranks per node then
  contend for the same injection path, so per-message overhead is paid several times
  over.

The fix is to make the message count depend on the number of *peer ranks* — a small,
topology-bounded number — instead of the number of *block-neighbour pairs*.

| | messages posted per exchange, per rank |
|---|---|
| before | number of off-rank (MeshBlock, neighbour) pairs — up to `56·M` |
| after | number of distinct peer ranks — bounded by the neighbourhood of the rank's sub-domain |

On-rank neighbours are untouched: they keep the existing direct device-to-device write
into the destination's receive buffer, with no MPI involved.

---

### 2.1 Cached metadata

`MeshBoundaryValues::BuildRankPackedVarMetadata(nvars)` (in `src/bvals/bvals.cpp`) walks
every `(m, n)` block-neighbour pair once and, for each off-rank neighbour, records a
`RankPackedVarEntry`:

```
struct RankPackedVarEntry { int m, n, lid, dn, data_size, offset; };
```

Entries are grouped by destination rank in a `std::map<int, …>`, which gives a
**deterministic, rank-ordered** layout on both sides of every exchange. Each group
becomes one `RankPackedVarMessage` describing a contiguous span of the aggregate buffer:

```
struct RankPackedVarMessage { int rank, nentries, entry_offset, hdr_offset, offset, data_size; };
```

Two flat device buffers, `rank_sendbuf_vars_` and `rank_recvbuf_vars_`, are sized to the
totals and hold every off-rank payload for the whole rank.

The metadata is rebuilt only when it can actually have changed — see §2.4.

### 2.2 One-shot header exchange, payload-only steps

A receiver must know where each incoming payload belongs. Rather than shipping that
layout on every step, it is exchanged **once**, when the metadata is built: each rank
sends its `(lid, dn, data_size)` triples to each peer under `tag = 2`, and uses what it
receives to build the offset maps below.

After that, the per-step path carries **payload only** (`tag = 1`). There is no
per-step header message, no host-side header rebuild, and no header request array.

Received headers are range-checked (`lid`, `dn`, `data_size`) before use, and a
malformed header is a fatal error rather than a silent out-of-bounds write.

### 2.3 Fused pack and unpack — no extra kernels, no extra fences

The obvious implementation — pack into the existing per-buffer `sbuf[n].vars`, then run
a second kernel to gather those into the aggregate buffer — costs two extra kernel
launches and two extra device fences per exchange. At small block sizes that overhead
exceeds the messaging win, and the aggregated path *loses* to the per-pair path.

Instead the aggregation is folded into the kernels that already exist. Two device maps,

```
DvceArray1D<int> send_agg_offset_, recv_agg_offset_;   // dimensioned (nmb*nnghbr)
```

hold, for each `(m, n)`, the base offset of that payload inside the aggregate buffer
(`-1` for on-rank or non-existent neighbours). With those in hand:

- `SendBuff` writes off-rank data straight to `aggsbuf(base + bufidx)` instead of
  `sbuf[n].vars(m, bufidx)`;
- `RecvBuff` reads off-rank data straight from `aggrbuf(base + bufidx)` instead of
  `rbuf[n].vars(m, bufidx)`.

So the aggregated path uses **exactly the same number of kernel launches and fences as
the original per-pair path**. This is what makes the change a win at *every* block size
rather than only at large ones.

All aggregate-buffer paths are guarded by `#if MPI_PARALLEL_ENABLED`, with the original
`sbuf`/`rbuf` expressions retained under `#else`, so serial builds are unaffected.

### 2.4 Invalidation under AMR

Cached metadata is valid until the mesh topology or the rank↔block mapping changes.
`Mesh` gains a monotonic counter:

```cpp
void MarkMeshUpdated() { ++amr_lb_seq_; }
int  GetAMRLoadBalanceUpdateSeq() const { return amr_lb_seq_; }
```

incremented once in `MeshRefinement::AdaptiveMeshRefinement` after
`RedistAndRefineMeshBlocks` — covering both refinement and any load balancing it
triggers. `InitRecv`, `PackAndSendCC` and `PackAndSendFC` rebuild when either `nvars` or
the sequence number differs from the cached value.

A monotonic counter is used rather than a boolean "mesh was updated this cycle" flag so
that the check is independent of *when* in the cycle it is made, and cannot be missed by
a consumer that runs after the flag would have been cleared.

---

## 3. Files changed

| file | change |
|---|---|
| `src/bvals/bvals.hpp` | `RankPackedVarEntry` / `RankPackedVarMessage`; cached metadata, aggregate buffers and offset maps; declarations of `GetVarDataSize` and `BuildRankPackedVarMetadata` |
| `src/bvals/bvals.cpp` | `GetVarDataSize`; `BuildRankPackedVarMetadata` (entry/message tables, buffer sizing, one-shot header exchange, offset maps) |
| `src/bvals/bvals_cc.cpp` | `PackAndSendCC` / `RecvAndUnpackCC`: fused aggregate write/read, per-peer `Isend` |
| `src/bvals/bvals_fc.cpp` | `PackAndSendFC` / `RecvAndUnpackFC`: same for face-centred fields |
| `src/bvals/bvals_tasks.cpp` | `InitRecv` posts per-peer `Irecv`; `ClearRecv`/`ClearSend` wait on the per-peer requests |
| `src/mesh/mesh.hpp` | `MarkMeshUpdated()` / `GetAMRLoadBalanceUpdateSeq()` and the `amr_lb_seq_` counter |
| `src/mesh/mesh_refinement.cpp` | one `MarkMeshUpdated()` call after `RedistAndRefineMeshBlocks` |

No new input parameters, no runtime toggle: the rank-packed path is the path, for both
cell-centred and face-centred variables. Flux-correction communication is unchanged.

---

### 4.1 CPU + MPI, against `main`

MHD linear wave (`inputs/tests/linear_wave_mhd*.athinput`, exercises both the
cell-centred and face-centred paths), Z4c linear wave (exercises the wider
`isame_z4c` buffers). Reference binary built from a pristine checkout of `main` in a
separate worktree; comparison is a byte-compare of `*-errs.dat` / `*.hst`.

| test | 1 rank | 2 ranks | 4 ranks |
|---|---|---|---|
| MHD linwave, uniform, 16 blocks | identical | identical | identical |
| MHD linwave, uniform, 128 blocks | identical | identical | identical |
| MHD linwave, SMR | identical | identical | identical |
| MHD linwave, AMR | identical | identical | see below |
| Z4c linwave, 64 blocks | identical | identical | identical |

The rank-packed build is also **decomposition-invariant**: the 1-rank and 4-rank error
norms are byte-identical to each other at both block counts.

At **AMR on 4 ranks** the norms agree to every printed digit except occasional
last-digit motion in a single norm (`M3_L1` = `1.898073e-04` vs `1.898074e-04`;
`d_L1` likewise moved by one unit in the last printed digit in one run out of three).
The cycle count, `L-infty`, and all remaining norms are identical. This is
non-associative reordering of the parallel reductions after a refinement event, not a
change of state: the same behaviour is seen on GPU (§4.2), and the difference does not
grow with time.

### 4.2 GPU, Vista (NVIDIA B200)

Orszag-Tang MHD, 4 GPUs on one node, rank-packed vs the identical build without it:

- **Uniform mesh: bit-identical.** History file identical over all 1640 cycles to
  t = 0.5; final binary dump identical to the last bit for all eight variables
  (`dens`, `vel{x,y,z}`, `eint`, `bcc{1,2,3}`), every cell, at 16, 64 and 256
  MeshBlocks.
- **AMR: agrees to round-off, and the difference does not grow.** History file
  bit-identical through t ≈ 0.45, after which differences appear at the 1e-10 level and
  then fluctuate between 1e-16 and 2e-10 — including dropping back to 5.5e-16 — rather
  than growing. Both variants execute identical MeshBlock-cycle counts, so refinement
  is unchanged. If this were a real error, the chaotic Orszag-Tang dynamics would
  amplify it to O(1).

### 4.3 Regression suite

`tst/run_test_suite.py --mpicpu` (29 tests) and `--style`.

---

### 5.1 Weak scaling, Vista B200

3D hydro linear wave, PLM / LLF / RK2, 256³ zones per GPU held fixed while the mesh and
GPU count grow together. The only difference between the two builds is the boundary
communication path; the reconstruction and Riemann-solver path is identical in both.
`N = 1` has no off-rank neighbours and is therefore a null control.

**Speedup (with rank-packed ÷ without):**

| GPUs | 1 blk/GPU (256³) | 8 (128³) | 64 (64³) | 512 (32³) | 4096 (16³) |
|---:|---:|---:|---:|---:|---:|
| 1 | 1.01x | 1.00x | 1.01x | 1.00x | 1.06x |
| 2 | 1.04x | 1.05x | 1.37x | 1.97x | 3.13x |
| 4 | 1.06x | 1.11x | 1.66x | 3.10x | 6.37x |
| 8 | 0.99x | 1.08x | 1.45x | 3.45x | 6.76x |

The more informative view is what it does to weak-scaling efficiency. Per-GPU
throughput at 8 GPUs relative to 1 GPU:

| blocks/GPU | 1 | 8 | 64 | 512 | 4096 |
|---|---:|---:|---:|---:|---:|
| without rank-packed | 0.92 | 0.84 | 0.63 | 0.27 | **0.15** |
| with rank-packed | 0.90 | 0.91 | 0.90 | 0.93 | **0.95** |

Without rank-packing, weak scaling degrades sharply as blocks get smaller — at 4096
blocks per GPU only 15% of single-GPU throughput survives to 8 GPUs. With it,
efficiency is 0.90–0.95 essentially independently of block size. The change does not
just make the code faster; it removes block size as a scaling variable.

### 5.2 Orszag-Tang MHD, Vista B200, 512², 4 GPUs

Aggregate zone-cycles/s, uniform mesh (exercises both CC and FC paths):

| MeshBlocks (per GPU) | block size | with | without | speedup |
|---|---|---:|---:|---:|
| 16 (4/GPU) | 128² | 1.676e8 | 1.161e8 | 1.44x |
| 64 (16/GPU) | 64² | 1.566e8 | 1.065e8 | 1.47x |
| 256 (64/GPU) | 32² | 1.473e8 | 0.968e8 | 1.52x |

### 5.3 Weak scaling with AMR, Vista B200

3D hydro linear wave, 256³ zones/GPU at root, `num_levels = 2`,
`refinement_interval = 5`, `min_max` criterion on `hydro_w_d` tracking the wave crest.
Final MeshBlock counts are identical between the two builds (refinement is
deterministic). Speedup (with ÷ without):

| GPUs | 8 blk/GPU | 64 | 512 | 4096 |
|---:|---:|---:|---:|---:|
| 1 | 1.00 | 1.00 | 1.01 | 1.04 |
| 2 | 1.07 | 1.27 | 1.91 | 3.18 |
| 4 | 1.15 | 1.31 | 2.24 | 4.21 |
| 8 | 1.10 | 1.25 | 2.15 | 5.89 |
| 16 | 1.05 | 1.26 | 2.51 | — |

### 5.4 Known regime where this is not a win

With **refinement every cycle** (`refinement_interval = 1`) the cached metadata is
rebuilt every cycle, and the rebuild — including its one-shot header exchange —
is not amortised. Orszag-Tang AMR on 4 B200 GPUs:

| configuration | with | without | ratio |
|---|---:|---:|---:|
| uniform | 1.676e8 | 1.161e8 | 1.44x |
| AMR, `refinement_interval = 1` | 3.688e7 | 4.724e7 | 0.78x |
| AMR, `refinement_interval = 10` | 6.530e7 | 6.909e7 | 0.95x |

The deficit also narrows as the root block count grows (0.78x at 64 root blocks,
0.90x at 256), and is gone by `refinement_interval = 5` in the weak-scaling study
above, where the change is a 1.05–5.9x win.

`refinement_interval = 1` is not a common production setting, but the cost is real and
the fix is known: make the rebuild incremental, so that only blocks whose neighbour
set actually changed are re-derived, and skip the header re-exchange for peers whose
entry list is unchanged. That is left for a follow-up rather than folded into this
change.

---

## 6. Notes for reviewers

- **No new input parameters and no runtime toggle.** The rank-packed path replaces the
  per-pair path for both CC and FC variables rather than sitting behind a flag.
- **Serial builds are unaffected.** Every aggregate-buffer path is under
  `#if MPI_PARALLEL_ENABLED` with the original expression under `#else`.
- **Buffer sizing.** `rank_{send,recv}buf_vars_` are sized from the summed entry
  `data_size`, which comes from the same `isame/icoar/ifine` index bookkeeping the
  per-pair path uses, via `GetVarDataSize`. The `is_z4c_` case selects
  `isame_z4c_ndat` exactly as before.
- **Ordering agreement between peers** rests on both sides iterating the same
  `std::map` key order (destination rank) and on the one-shot header exchange
  communicating the sender's within-message entry order. Neither side infers the
  other's layout.
- **Flux correction is unchanged** — `flux_correct_cc.cpp` and `flux_correct_fc.cpp`
  still use the per-pair path. Those exchanges are far less frequent, so aggregating
  them is a separate question.

### DIFF
--- a/src/bvals/bvals.cpp
+++ b/src/bvals/bvals.cpp
@@ -11,6 +11,8 @@
 #include <iostream>
 #include <utility>
 #include <algorithm> // max
+#include <map>
+#include <vector>
 
 #include "athena.hpp"
 #include "globals.hpp"
@@ -26,7 +28,19 @@
 MeshBoundaryValues::MeshBoundaryValues(MeshBlockPack *pp, ParameterInput *pin, bool z4c) :
   u_in("uin",1,1),
   b_in("bin",1,1),
-  i_in("iin",1,1),
+  i_in("iin",1,1)
+#if MPI_PARALLEL_ENABLED
+  ,
+  rank_packed_bvals_nvars_(-1),
+  rank_packed_mesh_seq_(-1),
+  rank_sendbuf_vars_("rank_sendbuf_vars",1),
+  rank_recvbuf_vars_("rank_recvbuf_vars",1),
+  rank_sendhdr_vars_("rank_sendhdr_vars",1),
+  rank_recvhdr_vars_("rank_recvhdr_vars",1),
+  send_agg_offset_("send_agg_offset",1),
+  recv_agg_offset_("recv_agg_offset",1)
+#endif
+  ,
   pmy_pack(pp),
   is_z4c_(z4c) {
   // allocate vector of status flags and MPI requests (if needed)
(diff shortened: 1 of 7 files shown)


jmstone (approved):

jmstone:
Thanks, David.  A major improvement to the code which I think will help a lot of people on a lot of different machines.

PR #776, opened 2026-07-30
Feature/multigrid

davidvelasco07:
# Multigrid self-gravity solver

A performance-portable geometric multigrid Poisson solver for self-gravity, with
full-multigrid (FMG) and iterative V-cycle drivers, SMR/AMR support, multipole outer
boundary conditions, and self-gravity source terms for both hydro and MHD.

This PR builds on the rank-packed boundary-communication change (#775), which is already
in `main`. That dependency is functional, not just packaging: without rank-packed messaging
the solver does not scale past a couple of GPUs. Section 2 explains why and Section 4
quantifies it.

---

## 1. What the solver does

Self-gravity requires solving ∇²Φ = 4πGρ to convergence *every* time step — and, on this
branch, at every RK stage, so that the potential stays consistent with the density it is
paired with. Geometric multigrid is the standard way to do that at O(N) cost.

Implemented:

- **Two drivers.** Full multigrid (`full_multigrid = true`): one nested-iteration ramp
  from the coarsest level up. Iterative V-cycles (`full_multigrid = false`): either to a
  residual threshold (`threshold`) or a fixed count (`niteration`, selected with
  `threshold < 0`).
- **FAS formulation** with configurable pre/post smoothing (`npresmooth`,
  `npostsmooth`, `omega`).
- **Mesh hierarchy in three regimes** — per-MeshBlock levels down to 1³, an *octet*
  level below that (where blocks no longer tile the domain), and a root grid that can
  live on the host or on the device (`root_on_host`).
- **SMR and AMR**, including flux-conserving prolongation across level jumps and correct
  defect-norm reduction on the final ascent.
- **Boundary conditions**: periodic, zero-gradient, zero-fixed, and a multipole expansion
  for isolated systems (`mg_bc`, `mporder`, `mporigin`/`auto_mporigin`, `nodipole`).
- **Source terms** coupling Φ back into the hydro and MHD momentum and energy equations.

New problem generators: `jeans_wave` (linear Jeans dispersion), `binary_gravity` (two
isolated spheres, SMR), `be_collapse` (Bonnor-Ebert collapse, AMR).

---

## 2. Why this needs rank-packed communication

Multigrid is unusually hostile to per-message-latency-bound communication, for two
reasons that compound.

**Every level of every V-cycle exchanges boundaries.** A hydro step exchanges ghost zones
a fixed number of times per RK stage. A V-cycle exchanges them on the way down (restrict),
on the way up (prolongate), and around *every* smoothing sub-step at *every* level. With
`npresmooth = npostsmooth = 1` a single V-cycle over an 8-level hierarchy issues on the
order of a hundred separate exchange phases, and a time step runs several V-cycles.

**The messages get smaller at exactly the rate the count stays fixed.** Coarsening halves
the linear dimension per level, so a face buffer shrinks by 4× per level while the number
of (MeshBlock, neighbour) pairs — and therefore the number of messages under the per-pair
scheme — does not change at all. By the time the hierarchy reaches 2³ blocks, a face
buffer is a handful of doubles wrapped in a full MPI message. The coarse levels do almost
no arithmetic and almost all of the messaging.

The consequence is that the *solver* is not what limits scaling; the *message count* is.
Measured on Vista B200, weak scaling at 16 GPUs relative to 1 GPU, 256³ zones per GPU:

| MeshBlocks per GPU | 1 (256³) | 8 (128³) | 64 (64³) | 512 (32³) | 4096 (16³) |
|---|---:|---:|---:|---:|---:|
| per-pair MG messages | 0.58 | 0.21 | 0.06 | 0.02 | **0.01** |
| rank-packed MG messages | 0.67 | 0.63 | 0.64 | 0.70 | **0.81** |

With per-pair messaging the solver retains 1% of its single-GPU per-GPU throughput at
16 GPUs on a well-decomposed mesh. It is not a slow solver at that point; it is a
non-scaling one. Rank-packing the messages — one aggregated message per peer rank per
exchange — recovers 0.63–0.81 efficiency across the whole range.

This is why the MG comm path on this branch is unconditionally rank-packed, with no
runtime toggle, and why `src/multigrid/multigrid.hpp` includes `bvals/bvals.hpp` and
reuses `RankPackedVarEntry` / `RankPackedVarMessage` directly rather than duplicating
them.

### 2.1 One MG-specific addition: the per-level metadata cache

Rank-packed metadata is derived per (level, nvars, skip_fc). A V-cycle revisits each
level about a dozen times (pre-smooth, restrict, prolongate, post-smooth) and a time step
runs several V-cycles, so a single-slot metadata cache keyed on the *current* level is
evicted on every level transition — roughly 85 rebuilds per cycle on an 8-level hierarchy,
each doing several device allocations and host-to-device copies.

At small block sizes that churn dominated, and plain rank-packing was actually *slower*
than the per-pair path (0.73–0.80× at 1 MeshBlock/GPU). Caching each level's comm state
in `mg_rp_cache_[level]` — messages, requests, aggregate buffers and offset maps, written
back on level switch so in-flight requests stay paired with their buffers, invalidated on
the mesh-update sequence — cuts rebuilds from ~85 per cycle to one per level for the whole
run, and flips that regime to 1.19–1.24×. It also gains 2–3× over uncached rank-packing
across *all* block sizes.

Three-way comparison (per-GPU throughput, 16 GPUs, ×10⁸ zone-cycles/s):

| MeshBlocks per GPU | 1 | 8 | 64 | 512 | 4096 |
|---|---:|---:|---:|---:|---:|
| per-pair | 0.85 | 0.59 | 0.17 | 0.04 | 0.01 |
| rank-packed, no cache | 0.40 | 0.58 | 0.48 | 0.58 | 0.44 |
| rank-packed + per-level cache | **0.97** | **1.70** | **1.95** | **1.67** | **0.93** |

---

## 3. Scope

48 files, +12074/−33 against `main`.

| area | files | lines | contents |
|---|---:|---|---|
| `src/multigrid/` | 5 | +7647 | solver, comm layer, driver, task list |
| `src/gravity/` | 4 | +638 | `Gravity` physics module, MG-Poisson coupling |
| `src/pgen/tests/` | 3 | +1354 | `jeans_wave`, `binary_gravity`, `be_collapse` |
| `src/mesh/`, `src/driver/` | 3 | +21/−2 | `Gravity` module hookup, per-RK-stage solve |
| `src/bvals/bvals.hpp` | 1 | +2/−1 | the `mg_zerograd` / `mg_zerofixed` / `mg_multipole` boundary flags |
| `src/srcterms/`, `src/tasklist/`, `src/outputs/` | 5 | +137/−4 | self-gravity source terms, `grav_phi` output |
| `tst/test_suite/multigrid/` | 11 | +1361 | 3 tests × {cpu, gpu, mpicpu} + helpers |
| `inputs/`, `tst/inputs/` | 9 | +795 | test and example inputs |
| `vis/python/plot_slice.py` | 1 | +66/−10 | slice plotting for potential fields |

---

## 4. Performance

MG self-gravity weak scaling on Vista B200, 256³ zones per GPU held fixed while mesh and
GPU count grow together. `full_multigrid = false`, `threshold = -1`, `niteration = 3`.
Identical solver in every series — only the communication path differs.

**Speedup, rank-packed + per-level cache vs per-pair MG messages:**

| GPUs | 1 blk/GPU (256³) | 8 (128³) | 64 (64³) | 512 (32³) | 4096 (16³) |
|---:|---:|---:|---:|---:|---:|
| 1 | 1.0x | 1.0x | 1.1x | 1.2x | 1.5x |
| 2 | 1.2x | 2.3x | 5.8x | 15.1x | 26.9x |
| 4 | 1.3x | 3.4x | 12.7x | 37.6x | 71.8x |
| 8 | 1.1x | 3.1x | 12.4x | 42.7x | 100.7x |
| 16 | 1.1x | 2.9x | 11.7x | 44.2x | **116.6x** |

At 1 GPU there is no off-rank neighbour, so that row is the null control and the ~1.0–1.5×
is measurement noise plus the on-rank path. Everywhere else the speedup grows with *both*
GPU count and decomposition, which is the signature of a message-count bottleneck rather
than a bandwidth one. There is no regime in which the rank-packed path loses.

Repeating the sweep with `root_on_host = true` changes the ratios by only a few percent
(e.g. 102× vs 117× at 16 GPUs, 4096 blocks/GPU), so root-grid placement is not a
meaningful factor in this comparison.

**AMR.** Bonnor-Ebert collapse (`be_collapse`, adaptive, 5 levels), aggregate
zone-cycles/s:

| GPUs | per-pair | rank-packed | ratio |
|---:|---:|---:|---:|
| 1 | 3.07e6 | 2.97e6 | 1.03x |
| 2 | 0.52e6 | 2.79e6 | **5.4x** |
| 4 | 0.50e6 | 3.12e6 | **6.3x** |
| 8 | 0.86e6 | 3.65e6 | **4.3x** |

The per-pair path flatlines at ~0.5–0.9e6 for *any* multi-GPU configuration — it does not
scale at all — while rank-packed climbs from 2.79e6 to 3.65e6. AMR produces many small
blocks, which is precisely the regime where per-pair messaging collapses, so this is the
expected result on a real adaptive workload rather than a synthetic sweep.

---

### 5.1 Test suite

`tst/test_suite/multigrid/` adds three tests, each in CPU, GPU and MPI-CPU variants:

- **`poisson3d`** — defect convergence. Residual falls to the 1e-8 threshold within ~10
  V-cycles with an average per-cycle convergence ratio ≤ 0.07. Also checks
  **decomposition independence**: the final defect agrees across MeshBlock layouts to
  < 1e-4, which is the direct test of the boundary-communication path.
- **`jeans3d`** — physics. Measured growth rate (unstable) and oscillation frequency
  (stable) are compared against the analytic Jeans dispersion relation
  ω² = k²c_s²(1 − n_jeans²); relative error ≤ 1% stable, ≤ 3% unstable. Run through both
  the FMG and the iterative V-cycle driver.
- **`binary_gravity`** — two isolated spheres under static mesh refinement with multipole
  outer BCs, exercising the coarse/fine boundary machinery and the isolated-BC path.
  Asserted on solver convergence and on defect consistency across decompositions.

mg_validation

**Left:** V-cycle defect convergence for the same Poisson problem under nine different
domain decompositions — 8³, 16³ and 32³ MeshBlocks × 1, 2 and 4 MPI ranks. The curves lie
on top of one another for eight orders of magnitude, separating only once they reach the
discretisation floor near 1e-10, where the residual is numerical noise and its exact value
is reduction-order dependent. Mean per-cycle reduction 0.141, i.e. the 1e-8 threshold is
crossed in six V-cycles.

**Right:** Jeans dispersion. Measured |ω| against the analytic
ω² = k²c_s²(1 − n_J²) across both the stable (oscillating, n_J < 1) and unstable
(growing, n_J > 1) branches. Inset: relative error at 32³ and 64³, whose mean ratio of
0.26 matches the 0.25 expected of a second-order scheme. Errors are largest near n_J = 1,
where ω → 0 and the relative measure is ill-conditioned.

### 5.2 Rank-packed MG comms produce identical results

Validated against the legacy per-pair MG path by comparing the
printed per-iteration defect sequences:

- **Uniform**, Jeans self-gravity, 2 GPUs: defect sequences bit-identical, every digit.
- **SMR**, `binary_gravity`, 4 levels, 2 GPUs: defects bit-identical
  (2.729025e6 → … → 4.025687e-1 at iteration 6).
- **AMR**, `be_collapse`, 5 levels, 2 GPUs: all 80 defect lines bit-identical over 22
  cycles / ~4 regrids; extended to 102 cycles, **936 defect lines bit-identical**. This
  is the test that the per-level cache invalidates correctly on regrid — no stale
  buffers, offsets, or in-flight requests survive a mesh change.

### 5.3 Physical sanity

`binary_gravity` under SMR resolves the two separate potential wells (deeper massive
sphere, shallower companion) that a uniform 128³ mesh merges into one, the two centres
being ~1.5 uniform-mesh cells apart. Potential range and periodic zero-mean behaviour are
as expected.

---

## 6. Known limitations and follow-ups

- **Per-stage solve cost.** The driver solves gravity at every RK stage, so that the
  potential is consistent with the density it is paired with. Cost relative to one MHD
  step, weak scaling at 256³ zones/GPU on GH200 (isothermal MHD):

  | FMG cycle / MHD step | 1 GPU | 2 GPUs | 4 GPUs | 8 GPUs |
  |---|---:|---:|---:|---:|
  | 8 blk/GPU (128³) | 0.94 | 1.08 | 1.08 | 1.13 |
  | 512 blk/GPU (32³) | 0.37 | 0.43 | 0.43 | 0.45 |
  | 4096 blk/GPU (16³) | 0.35 | 0.36 | 0.35 | 0.35 |

  | V-cycle / MHD step | 1 GPU | 2 GPUs | 4 GPUs | 8 GPUs |
  |---|---:|---:|---:|---:|
  | 8 blk/GPU (128³) | 0.33 | 0.38 | 0.38 | 0.40 |
  | 512 blk/GPU (32³) | 0.17 | 0.19 | 0.19 | 0.20 |
  | 4096 blk/GPU (16³) | 0.16 | 0.16 | 0.16 | 0.15 |

  The rows are flat in GPU count: both MG and MHD grow ~20% in absolute time from 1→8
  GPUs, but together, so the ratio is GPU-count-independent. Self-gravity therefore costs
  a fixed fraction of the step no matter how many ranks the problem is spread over — which
  is a direct consequence of §2, since the rows only stay flat while the MG comm path
  scales.

  Cost falls as decomposition rises because MG is nearly decomposition-insensitive while
  MHD is not: MHD carries `nghost = 4` (a 24³ halo around a 16³ block, 3.4× overhead)
  against `mg_nghost = 1` (1.4×). The unfavourable corner is small meshes with few large
  blocks — across the full single-GPU mesh × block sweep FMG/MHD averages 2.19× and
  reaches 6.08× at 64³ with 8 blocks, though those runs are ~4–5 ms and timer-noise
  limited.
- **`gravity/mg_nghost > 1` is not usable with refinement.** The parameter is accepted and
  does take effect (`ngh_` sizes the MG arrays as `(nx >> ll) + 2*ngh_`), but the
  coarse/fine machinery assumes a single ghost layer.

- **FC prolongation transverse gradients are degraded on the coarsest block levels.**
  `ProlongateFCMG` (`multigrid_bvals.cpp`) clamps the transverse gradient stencil to stay
  inside the coarse buffer. With `half = ncells/2`, the clamp is
  `sjp = (sj < ngh+half-1) ? sj+1 : sj`, so at `ncells = 2` (`half = 1`) the two indices
  collapse onto the same cell and the gradients are **identically zero**; at `ncells = 4`
  every row is one-sided; centred gradients only appear from `ncells = 8` upward, where
  just the two edge rows stay one-sided. Below `ncells = 2` the routine returns early.

---


jmstone:
This is a significant new capability!  The description suggests that the rank-pack boundary exchange is an option.  Is it?  Section 5.2 suggest the toggle was removed.  I find Claude descriptions contain TMI and can be confusing.

Kengo's paper focused on "ghost forces" at fine/coarse boundaries?  Did you explore if this is an issue?

davidvelasco07:
Yes, Kengo's "mass conserving" interpolation was implemented to prevent that issue.
Here is an example of the "mass conserving" interpolation (on the left) against a direct interpolation (on the right)
image

PR #777, opened 2026-08-03
Refresh MHD ghosts and guard linear-wave neighbor reads

dfielding14:
## Summary

Follow up on #739 with two focused fixes:

- Rerun the standard boundary-value and primitive initialization after `RepairAMRFC()` so repaired face-centered magnetic fields are propagated to neighboring ghosts before reconstruction.
- Guard the linear-wave AMR vector-potential neighbor checks by dimensionality. 1D and 2D MeshBlocks allocate 8 and 24 neighbor entries, respectively, but the problem generator could read 2D/3D-only entries. Those out-of-bounds reads made the regenerated MHD reference state rank-layout dependent, first appearing in `B3`.

Both fixes use the existing communication and initialization paths; no new flags or boundary messages are needed. The existing GR linear-wave error bounds are unchanged.

## Why this surfaced after #739

This was a pre-existing problem-generator bug, not a GR/WENOZ solver regression. The invalid accesses date to 2022, and #739 did not modify `linear_wave.cpp`.

In 2D, each MeshBlock has 24 neighbor entries (indices 0–23), but the linear-wave AMR vector-potential correction also inspected 3D-only entries 24–47. With the array’s `LayoutRight` storage, these accesses can alias the next local MeshBlock’s neighbor row, while the final local block reads beyond the allocation. `LinearWaveErrors()` reruns the problem generator on the final AMR topology to construct the reference solution, so the measured error could depend on MeshBlock packing and memory state.

Consistent with this, the exact same #739 commit and source tree passed one 16-rank MPI CI run and failed another; the failure exceeded the bound by only 0.0145%. The AMR changes exposed this latent undefined behavior but did not introduce it. The dimensional guards remove the invalid reads, and the original regression bounds are unchanged.

## Validation

- MPI 2D and deep-3D-L2 div(B) AMR regressions
- Bit-for-bit identical 4-rank and 16-rank 2D GRMHD states through the first AMR regrid
- Full MPI 2D GR linear-wave suite: 4 passed
- MHD/RK3/WENOZ high-resolution RMS-L1 `1.994064e-06` (limit `2.0e-06`), convergence ratio `0.214408` (limit `0.22`)
- Python and C++ style checks: 2 passed

### DIFF
--- a/src/mesh/mesh_refinement.cpp
+++ b/src/mesh/mesh_refinement.cpp
@@ -163,14 +163,10 @@ void MeshRefinement::AdaptiveMeshRefinement(Driver *pdriver, ParameterInput *pin
     MeshBlockPack* pmbp = pmy_mesh->pmb_pack;
     if (pmbp->pmhd != nullptr) {
       RepairAMRFC(pmbp->pmhd->b0);
-      if (pmbp->pdyngr == nullptr) {
-        (void) pmbp->pmhd->ConToPrim(pdriver, 0);
-      } else {
-        if (pmbp->pz4c != nullptr) {
-          (void) pmbp->pz4c->ConvertZ4cToADM(pdriver, 0);
-        }
-        (void) pmbp->pdyngr->ConToPrim(pdriver, 0);
-      }
+      // Repair changes internal faces after the first exchange finalized exterior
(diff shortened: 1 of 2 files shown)


jmstone:
Good catch, thanks for these important updates.

PR #781, opened 2026-08-07
Extend the TOV pgen to use external dipole fields

spacetimecurv:
For a separate project I seeded a TOV in `dyngr_tov.cpp` with an external magnetic field that uses the same current-loop model as the `Elliptica` or `Lorene` pgens. Since this is a useful addition, I open this PR which does minimal changes and uses a struct to hold the necessary information about both configurations. The vector potential is then computed by the same function based on the specified magnetic field structure. This is short test I did for a TOV with a small velocity perturbation:
https://github.com/user-attachments/assets/afee5576-a36a-4c0c-a775-07e1fe674125

### DIFF
--- a/src/pgen/dyn_grmhd/dyngr_tov.cpp
+++ b/src/pgen/dyn_grmhd/dyngr_tov.cpp
@@ -12,6 +12,7 @@
 #include <iostream>
 #include <limits>
 #include <sstream>
+#include <string>
 #include <utility>
 
 #include "athena.hpp"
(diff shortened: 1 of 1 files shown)


jfields7 (approved):

jfields7:
If you could update [this page](https://ias-astrophysics.github.io/athenak-docs/dyngrmhd.html) in the documentation to reflect the new dipole parameters, that would be very helpful.

spacetimecurv:
> If you could update [this page](https://ias-astrophysics.github.io/athenak-docs/dyngrmhd.html) in the documentation to reflect the new dipole parameters, that would be very helpful.

Done, see https://github.com/IAS-Astrophysics/athenak-docs/pull/10.

PR #782, opened 2026-08-07
Eliminate subviews in AthenaTensor

jfields7:
This PR contains three changes:
1. `AthenaTensor` no longer uses `Kokkos::subview` to slice from a master array; instead, it contains a reference to the full data and uses an offset index do the mapping. In theory this is what `Kokkos::subview` does as well, but it contains significantly more metadata that inflates register usage substantially.
2. Rank-2 `AthenaTensor` objects no longer store an index map but instead access data similar to `AthenaPointTensor` via index arithmetic. This also reduces register usage.
3. The `z4c_onepuncture_amr.athinput` example input file was never updated after #667 and therefore failed to run. Now it does.

At least on my workstation, performance changes are not significant because there is still substantial register spillage in the RHS kernel. It is not likely that we will see significant performance improvements until we eliminate the spills altogether. Thus this PR should be considered a step in an ongoing project to improve Z4c performance. However, register spillage has been reduced significantly:
| Kernel | spill stores (old) | spill loads (old) | spill stores (new) | spill loads (new) | reduction in stores | reduction in spills |
| --------- | ---------------------- | --------------------- | ------------------------ | ---------------------- | -------------- | -------- |
| Z4c RHS, 2nd order | 2000 bytes | 2084 bytes | 316 bytes | 316 bytes | 84.2% | 84.8% |
| Z4c RHS, 4th order | 3404 bytes | 4876 bytes | 772 bytes | 1060 bytes |77.3% | 78.3% |
| Z4c RHS, 6th order | 1024 bytes | 1208 bytes | 232 bytes | 232 bytes | 77.3% | 80.8% |

Note that these numbers exclude changes made by #743, which added support for a slow-start lapse inside the Z4c RHS kernel and therefore changes register usage slightly. It's not clear to me why 6th order spills the least and 4th order spills the most, considering the 2nd order operators should require the fewest registers and 6th order the most, but I assume this is one of those places where the compiler is struggling to optimize the code well. A handful of other kernels, such as the 6th-order ADM constraints and the Weyl tensor calculation at all orders, also had minor spills before these changes, and they now fit completely inside the register space. There are also some minor benefits to the `DynGRMHD` flux and C2P kernels, which access the ADM variables via `AthenaTensor`. There are some other changes in the pipeline that will address those kernels more directly.

The changes are all behind the scenes inside `athena_tensor.hpp` and shouldn't affect anything anyone has been doing or require updates to the documentation.


HengruiZhu99 (approved):
Great! I will try adding my kernel splitting later to see if the two combined would improve the performance more significantly. The register spillage was a bit unexpected, on della, I was able to eliminate spillage by kernel splitting for the z4c rhs for the 2nd and 4th order, but not the 6th order. Will check again and see how the combined subview elimination and kernel splitting work. (Though definitely surprised that there's so much overhead for the subview...)

PR #785, opened 2026-08-13
Add angle grid options to SphericalSurface output class

spacetimecurv:
This PR extends the `SphericalSurface` output class with a set of grid options for the angles. So far, the grid in $\theta$ has used a grid uniform in $\cos{\theta}$ which has the downside, that the angular resolution degrades as we move to the poles. In some applications that might want to resolve polar outflows or else this might not be desirable. Also, with the old grid the node-centered points at the poles own half a cell but got a full weight which for surface integration can lead to significant errors. This is amplified by the fact that the points were spaced as $2/(n_{\theta}-1)$, while the weights assumed $2/n_{\theta}$. Here, we fix these problems and extend the class with some angular grid options:
1. In the old class we always set `ntheta` and `nphi` is then computed as `2*ntheta`. Per default this is also kept here, but if desired the user can now also independently choose `nphi`.
2. The parameter `theta_spacing` can be set to `uniform_costheta` (default; corresponds to the "old" grid; uniform in $\cos{\theta}$) and `uniform_theta` (the "new" grid; `dtheta=pi/ntheta` for cell or `pi/(ntheta-1)` for node).
3. The parameter `theta_centering` can be set to `node` (default; "old" option; first point on the z-axis) and `cell` ("new" option; first point is at `dtheta/2`).
4. `weights` is by default set to true (as in the old path) and will dump the weights to the corresponding .vtk files, but setting it to false would not write to file.

A PR has been opened for the updated documentation under https://github.com/IAS-Astrophysics/athenak-docs/pull/11.

### DIFF
--- a/src/outputs/outputs.hpp
+++ b/src/outputs/outputs.hpp
@@ -433,6 +433,7 @@ class SphericalSurfaceOutput : public BaseTypeOutput {
   void WriteOutputFile(Mesh *pm, ParameterInput *pin) override;
  private:
   SphericalSurface *psurf;
+  bool dump_weights;  // write the quadrature weight of each point to file
 };
 //----------------------------------------------------------------------------------------
 //! \class EventLogOutput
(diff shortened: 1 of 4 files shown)


dradice (approved):
This looks like a good feature to have. Note that a grid uniform in cos theta is as close as possible to equiarea for a latitude / longitude grid so it should be preferred in most applications and I would keep it as default (as it is in your patch).

jfields7 (approved):

PR #779, opened 2026-08-05
Add RKL2 super-time-stepping for diffusion

dfielding14:
## Summary

This PR adds optional second-order Runge-Kutta-Legendre super-time-stepping (RKL2 STS) for the existing constant-coefficient diffusion operators:

- isotropic thermal conduction in Newtonian Hydro and MHD
- isotropic viscosity in Newtonian Hydro and MHD
- Ohmic resistivity in Newtonian MHD, using constrained transport

Explicit integration remains the default, and each diffusion process independently selects explicit or STS integration.

When enabled, the STS update is applied as symmetric half-sweeps before and after the ordinary Runge-Kutta update. The implementation reuses AthenaK's existing task lists, boundary communication, flux correction, AMR, MPI, and constrained-transport machinery.

## Configuration

STS uses two levels of configuration.

First, the `<time>` block selects the STS algorithm available to the run:

```text
<time>
sts_integrator = rkl2
```

Second, each active diffusion process chooses whether it is advanced explicitly or with STS:

```text
conductivity_integrator = explicit  # or sts
viscosity_integrator = explicit     # or sts
resistivity_integrator = explicit   # or sts
```

Thus, setting `sts_integrator = rkl2` does not automatically move every diffusion process into the STS sweeps. It enables the RKL2 controller; the individual process settings determine which diffusion operators use it.

For example, this configuration advances thermal conduction with STS while leaving viscosity in the ordinary explicit update:

```text
<time>
sts_integrator = rkl2
sts_max_dt_ratio = 16.0

<hydro>
alpha_iso = 0.5
conductivity_integrator = sts

nu_iso = 0.25
viscosity_integrator = explicit
```

The explicit viscosity limit continues to constrain the ordinary timestep, while conduction is advanced in the two RKL2 half-sweeps.

Multiple processes can use STS simultaneously:

```text
<time>
sts_integrator = rkl2
sts_max_dt_ratio = 16.0

<hydro>
alpha_iso = 0.5
conductivity_integrator = sts

nu_iso = 0.25
viscosity_integrator = sts
```

In this case, the conduction and viscosity stability rates are combined when selecting the number of RKL2 stages, and both operators are advanced during the same STS sweeps.

For Ohmic diffusion in MHD, a minimal configuration is:

```text
<time>
sts_integrator = rkl2
sts_max_dt_ratio = 16.0

<mhd>
eta_ohm = 0.25
resistivity_integrator = sts
```

[...]

PR #787, opened 2026-08-18
fix(test): the `test_sts_cycle_boundary_restart` test was failing

bcaddy:
This test was failing for my MPI compiler since the MPI version of AthenaK wasn't being launched with MPI. Adding `mpirun` and using `testutils.mpi_run` fixed the issue for me. This test was introduced in PR #779.

@dfielding14, could you take a look to make sure I haven't broken this test?

### DIFF
--- a/tst/test_suite/diffusion/test_sts_rkl2_mpicpu.py
+++ b/tst/test_suite/diffusion/test_sts_rkl2_mpicpu.py
@@ -130,9 +130,10 @@ def test_sts_cycle_boundary_restart(tmp_path):
     shutil.rmtree("rst", ignore_errors=True)
     try:
         direct = "sts_restart_direct"
-        assert testutils.run(
+        assert testutils.mpi_run(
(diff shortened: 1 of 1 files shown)


dfielding14:
Thanks for catching this! Your fix looks right to me. The test is intentionally a one-rank restart-consistency check, and I had mistakenly relied on the MPI-enabled executable supporting singleton startup. Using `mpi_run(..., threads=1)` and `mpirun -np 1` for the restart preserves the original test exactly while making it portable across MPI implementations. CI is green, so this looks good to merge.

bcaddy:
Most MPI implementations I’ve come across support singleton startup but for some reason it doesn’t reliably work with NVHPC. Sometimes programs will launch just fine, other times not. They always work with mpirun though

jmstone:
Bob: thanks for catching this, and Drummond: thanks for confirming the fix.

